PR #15, opened 2019-02-26
Table transform endpoints

subdavis:
@jbeezley this is just for me to keep an eye on the diff because I like the github interface for it.

### DIFF
--- a/metabulo/models.py
+++ b/metabulo/models.py
@@ -5,13 +5,16 @@ from uuid import uuid4
 
 from flask import current_app
 from flask_sqlalchemy import SQLAlchemy
-from marshmallow import fields, post_dump, post_load, pre_load, Schema
+from marshmallow import fields, post_dump, post_load, Schema
 from marshmallow.exceptions import ValidationError
 import pandas
 from sqlalchemy import DDL, event, MetaData
 from sqlalchemy.schema import UniqueConstraint
+from sqlalchemy_utils.types.json import JSONType
 from sqlalchemy_utils.types.uuid import UUIDType
 from werkzeug.utils import secure_filename
+
+from metabulo import transform
 
 
 # This is to avoid having to manually name all constraints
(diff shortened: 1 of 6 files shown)


metabulo/models.py
```diff
@@ -121,36 +146,50 @@ def read_csv_file(self, data):
...
-    DDL("""
-INSERT INTO csv_column_type (type_, numeric) VALUES
-('object', 0), ('float64', 1)
-    """)
-)
+    TransformType.__table__, 'after_create',
+    DDL(f"""
+INSERT INTO transform_type (name) VALUES
+{','.join(["('" + t + "')" for t in transform.registry.keys()])}
+"""))
```
subdavis: Why does the registry need to be duplicated into the db?  Does this pr support adding new transforms after the initial create event?

Writing and maintaining migrations on stumpf wasn't too fun, so I'm curious if there are ways to avoid setting the project up to need those in the future.

jbeezley:
@subdavis This is in a usable state now.  There is a generic API for chaining multiple transforms on a file.

```
POST /csv/:id/transform
DELETE /csv/:id/transform/:id
```

The POST endpoint accepts a json body that is validated against a marshmallow schema specific to each transform, e.g.
```json
{
  "transform_type": "normalize",
  "prority": 1
}
```
For transforms taking an argument, they are added to the json body, e.g.
```json
{
  "transform_type": "set_value",
  "prority": 2,
  "column": "column name",
  "row": "row name",
  "value": 0
}
```

The response to the POST request also contains new version of the transformed csv_file model, so you shouldn't have to rerun the `GET /csv/:id` endpoint.

Let me know if you have problems.

metabulo/models.py
jbeezley: Oh yes.  For the moment, I'm avoiding doing migrations because we don't have any data we actually need to save.  Until we do, the migration procedure is delete the sqlite database and rerun `metabulo-create-tables`.

jbeezley: > Why does the registry need to be duplicated into the db? Does this pr support adding new transforms after the initial create event?

It doesn't really.  At the moment, it's a sqlite version of ENUM.  There may be a need to store persistent information about the registry at some point.  If not, it should be easy enough to delete it.  (I already deleted the similar column type table that had existed.)

PR #18, opened 2019-02-26
add plot

subdavis:
Add a generic looking box plot.


subdavis:
screenshot_20190226_141115

Here's where we are at the moment.

PR #19, opened 2019-02-26
Add an endpoint that generates a PCA overview plot

jbeezley:
This is almost certainly not what we want to do, but I'm adding it in service of doing something non-trivial with opencpu.

This takes 2.5 seconds to return the image, but there is a lot of overhead in the implementation.  Since I don't really know R, I had to run it through the full metaboanalyst pipeline, which writes all kinds of files out to disk for the web server.  I think it would be much faster as a simple in-memory implementation (read csv -> PCA -> scatter plot -> return png).


subdavis:
Can I merge it?

Issue #22, opened 2019-02-27
Create table cleanup interface


subdavis:
Should probably be very similar to the one in SIMCA.

I'm going to start a 30 day trial of it to get a feel for how this interface works, and try to break that into a detailed list of mvp features.

subdavis:
After a few minutes of playing with it, I found the data cleanup interface is actually a good bit more straightforward than I was expecting.

You can set primary and secondary keys for variables and observables.  You can flag data as qualitative, quantitative, or datetime.  The only thing I don't really get is the difference between a qualitative classification and a secondary ID classification.

There are popouts that show variables and observables, and you can toggle them on and off by clicking the rows.

I think it would be enough to just populate a select dropdown with:

* <index> (means quantitative data by default)
* primary
* secondary
* datetime
* enable/disable

Naturally the UX should go through Jared, but this is enough to get the job done until we have a better idea of what's really needed.

Screens:

![capture](https://user-images.githubusercontent.com/4214172/53585373-c0886780-3b53-11e9-89b1-b90afd95000b.PNG)
![capture2](https://user-images.githubusercontent.com/4214172/53585374-c0886780-3b53-11e9-8a4b-8eeed2000a00.PNG)
![capture3](https://user-images.githubusercontent.com/4214172/53585375-c0886780-3b53-11e9-9be5-d7ddb447e5ca.PNG)

subdavis:
@jbeezley I see in #25 you added types for row and column.  Awesome!

```python
TABLE_COLUMN_TYPES = ['primary-id', 'secondary-id', 'numeric', 'qualitative']
TABLE_ROW_TYPES = ['header', 'sample', 'other']
```

It seems that this, the UI, and SIMCA all conflict in small ways.

1. Mutual exclusion.  You and I assume that row/col can only have 1 type, but SIMCA doesn't.  It looks like that advanced behavior is related to the "Variable Roles" which we don't care about (yet).  Our assumption seems safe for the time being.
  
2. The backed uses primary and secondary for both row and col, but you have header and sample.  I don't know what the convention should be, and this may be for someone like Tom to decide.  It's extra work to make them different -- could we make them the same for now (unless you understand the difference, of course)?

Thanks!

jbeezley:
> Mutual exclusion...

[...]

PR #24, opened 2019-02-28
Cleanup Table

subdavis:
Starting MVP for a cleanup table for #22 

My goal for this particular widget is to use more native features and less Vuetify to keep the DOM lighter and more responsive.  For example, those are select dropdowns on the axes rather than `v-select` elements with non-native popovers.   The coloring priority is done with css.

We can, of course, try to incorporate more vuetify if the situation calls for it.  

screenshot_20190228_103303


jeffbaumes:
Great idea to keep things simple (non-Vuetify) so we can optimize for scalability and data screen density. When we know the indicators we need for the various cells, Jared could help with color and/or other cues for colorblind users, maybe icons. Maybe even drop color in favor of icons?

subdavis:
Tweaked a bit more.   Going to merge this in now to avoid conflicts if anyone else is working on UI stuff.

screenshot_20190301_160202
screenshot_20190301_160223
screenshot_20190301_160246

PR #31, opened 2019-03-05
Add PCA json endpoint

subdavis:
Returns `{x: Number, y: Number, label: String}`

### DIFF
--- a/metabulo/models.py
+++ b/metabulo/models.py
@@ -1,3 +1,4 @@
+from collections import namedtuple
 from datetime import datetime
 from io import BytesIO
 from pathlib import Path, PurePath
@@ -29,14 +30,9 @@ metadata = MetaData(naming_convention={
 })
 db = SQLAlchemy(metadata=metadata)
 
-TABLE_COLUMN_TYPES = ['key', 'metadata', 'data', 'masked']
-COLUMN_KEY_INDEX = 0
-COLUMN_METADATA_INDEX = 1
-COLUMN_DATA_INDEX = 2
-TABLE_ROW_TYPES = ['header', 'metadata', 'sample', 'masked']
-ROW_KEY_INDEX = 0
-ROW_METADATA_INDEX = 1
-ROW_DATA_INDEX = 2
+TableTypes = namedtuple('TableTypes', ['INDEX', 'METADATA', 'DATA', 'MASK'])
+TABLE_COLUMN_TYPES = TableTypes('key', 'metadata', 'metabolite', 'masked')
+TABLE_ROW_TYPES = TableTypes('header', 'metadata', 'sample', 'masked')
 
 
 class BaseSchema(Schema):
@@ -134,16 +130,16 @@ class CSVFile(db.Model):
                 'csv_file_id': self.id,
                 'column_header': table.index.name,
                 'column_index': index,
-                'column_type': TABLE_COLUMN_TYPES[COLUMN_KEY_INDEX],
+                'column_type': TABLE_COLUMN_TYPES.INDEX,
             })
         )
         index += 1
         for column_header, dtype in table.dtypes.items():
             # There are probably better heuristics for this.
             if dtype == np.object:
-                column_type = TABLE_COLUMN_TYPES[COLUMN_METADATA_INDEX]
+                column_type = TABLE_COLUMN_TYPES.METADATA
             else:
-                column_type = TABLE_COLUMN_TYPES[COLUMN_DATA_INDEX]
+                column_type = TABLE_COLUMN_TYPES.DATA
 
             columns.append(
                 table_column_schema.load({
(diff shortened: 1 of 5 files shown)


subdavis:
This is just a small change that 

* adds PCA, 
* adds constants to reference column types rather than raw strings, 
* and changes `metabolite` to `data` because I don't think anything on the backend is specialiazed for the domain of metabolomics yet.  We can also process binned NMR spectra.

jbeezley:
I'll probably rebase this on top of my new PR.  Some of the logic here can be simplified.  Once I'm done with this, you can take a crack at fixing the indexing

subdavis:
Sounds good to me.  Thanks.

jbeezley:
@subdavis PTAL at my changes.

metabulo/models.py
```diff
@@ -29,8 +30,9 @@
 })
 db = SQLAlchemy(metadata=metadata)
 
-TABLE_COLUMN_TYPES = ['key', 'metadata', 'metabolite', 'masked']
-TABLE_ROW_TYPES = ['header', 'metadata', 'sample', 'masked']
+TableTypes = namedtuple('TableTypes', ['INDEX', 'METADATA', 'DATA', 'MASK'])
+TABLE_COLUMN_TYPES = TableTypes('key', 'metadata', 'metabolite', 'masked')
+TABLE_ROW_TYPES = TableTypes('header', 'metadata', 'sample', 'masked')
```
subdavis: if `len(column types) != len(row types)` this is easy to fix right?  We can just break the named tuples into row and col.

subdavis: Could we trade `metabolite` and `metabolites` as strings and variable names for `measurement` and `measurements`?

I can do this in a separate PR, since it doesn't really affect this one.

metabulo/plot.py
```diff
@@ -17,3 +18,17 @@ def make_box_plot(frame):
     output = io.BytesIO()
     FigureCanvas(fig1).print_png(output)
     return output.getvalue()
+
+
+def pca(metabolites, metadata, x_component_index, y_component_index):
+    n_components = max(x_component_index, y_component_index) + 1
```
subdavis: is this the same as `n_components == min(n_samples, n_features)`  ?

metabulo/plot.py
```diff
@@ -17,3 +18,17 @@ def make_box_plot(frame):
...
+    components = PCA(n_components=n_components).fit_transform(metabolites)
+    samples = []
+    for i in range(metabolites.shape[0]):
+        samples.append({
+            'x': components[i][x_component_index],
+            'y': components[i][y_component_index],
+            'key': metabolites.index[i],
+            'labels': metadata.iloc[i].to_dict()
+        })
+    return samples
```
subdavis: What's your gut on writing a Marshmallow schema here to load this into?  I thought about it, but it seemed like overkill.

metabulo/models.py
jbeezley: Yes and sure.

metabulo/plot.py
jbeezley: No.  This is just more efficient because it only computes the minimal number of singular vectors (SVD is rather slow especially if we get into hundreds of metabolites).

With the code you wrote, it was always returning the largest principal component vs. the second largest.  In that case, you only needed the first two components (the two largest singular vectors).  I modified the endpoint to scatter plot the samples projected onto arbitrary components in each axis.

BTW if we ever do anything like the PCA overview plot from metaboanalyst, it will be cheaper just to return all combinations in one call returning a (ncomponents x ncomponents x nsamples) array.

metabulo/views.py
```diff
@@ -177,6 +177,17 @@ def get_box_plot(csv_id):
     return Response(fig, mimetype='image/png')
 
 
+@csv_bp.route('/csv/<uuid:csv_id>/plot/pca', methods=['GET'])
+def get_pca_plot(csv_id):
+    csv_file = CSVFile.query.get_or_404(csv_id)
+    x_component_index = int(request.args.get('x_component_index', 0))
+    y_component_index = int(request.args.get('y_component_index', 1))
```
subdavis: why would I want to change these in my request?  to flip the axis?

metabulo/plot.py
jbeezley: Seems overkill to me as well.  We could try webargs for this sort of thing.

metabulo/views.py
jbeezley: To choose which principal component is shown on each axis.  I think I remember Metaboanalyst having a feature like that.

subdavis: Edit: nevermind

metabulo/plot.py
subdavis: :+1:

PR #36, opened 2019-03-06
Generalize Primary Keys

subdavis:


subdavis:
Still working on this, but the branch is stable for the time being.

subdavis:
This branch is now more or less unusable, and is more of a personal sandbox.  I think something useful may come out of it.

Master should be functional now -- pushed some minor bugfixes this morning.  Those mostly stemmed from incompatibilities with the backend added when the label API changed.

jbeezley:
Sounds good.  I left this as a TODO in the code because I suspected it would be a minefield.  Sounds like I was right. :)

I may try to implement a reduced version of this early next week... That is, just allowing the user to change the column containing the row keys.

metabulo/models.py
```diff
@@ -248,28 +253,34 @@ def make_object(self, data):
...
     column_index = db.Column(db.Integer, primary_key=True)
-    column_header = db.Column(db.String, nullable=False)
     column_type = db.Column(db.String, nullable=False)
 
+    @property
+    def column_header(self):
+        if self.csv_file.index_row is not None:
+            index = self.csv_file.index_row.row_index
+            return self.csv_file.table.iat[index, self.column_index]
+        return ''
```
subdavis: @jbeezley yeah, it was a doozy.  My approach was to remove `column_header` and `row_name` from the data model and make them "computed" properties.   Then, when you change a key, it's no different than changing metadata or masks, which is super nice.

The Catch (TM)  is that this relies total removal of reliance on pandas' indexes.    I currently still believe that breaking with the pandas data model is inevitable, so we might as well get it over with, because it's just going to make life harder later.

There _is_ a hybrid approach where we reorder the array with the indexes in row/col0 when it's read from disk.  To borrow my new favorite phrase from Zach, that gave me bad tummy feels.

jbeezley: I still think we can avoid abandoning the pandas indexing without too many bad tummy feels, but I might just have to prove myself wrong when I attempt it.

subdavis:
Alright, it took some doing, but I think I this meets my goal.

I wanted to be able to properly process a file that looks like this:

```text
This one is just empty,This is a comment,,,,,,
,Something else,,,,,,
,Sample,Label,"1,6-Anhydro-beta-D-glucose",1-Methylnicotinamide,2-Aminobutyrate,2-Hydroxyisobutyrate,2-Oxoglutarate
,,,META1,MET2,META3,META4,META5
,PIF_178,cachexic,40.85,65.37,18.73,26.05,71.52
,PIF_087,cachexic,62.18,340.36,24.29,41.68,67.36
,PIF_090,cachexic,270.43,64.72,12.18,65.37,23.81
,NETL_005_V1,cachexic,154.47,52.98,172.43,74.44,1199.91
,PIF_115,cachexic,22.2,73.7,15.64,83.93,33.12
,PIF_110,cachexic,212.72,31.82,18.36,80.64,47.94
,,,,,,,
,NETCR_014_V1,cachexic,31.5,6.82,4.18,12.94,25.03
,NETCR_014_V2,cachexic,51.42,30.27,7.54,34.81,80.64
```

This branch makes guesses about what rows and columns should be data, metadata, keys, and which rows are invalid and should be masked.

Without input, it should reduce the above to:

```
40.85,65.37,18.73,26.05,71.52
62.18,340.36,24.29,41.68,67.36
270.43,64.72,12.18,65.37,23.81
154.47,52.98,172.43,74.44,1199.91
22.2,73.7,15.64,83.93,33.12
212.72,31.82,18.36,80.64,47.94
151.41,36.6,8.67,42.52,223.63
31.5,6.82,4.18,12.94,25.03
51.42,30.27,7.54,34.81,80.64
```

...with the key and header rows you would expect.

There are a few breaking changes in the API to get here (only really related to row/col label PUT).  I'm not going to fix the UI until other folks have had a chance to provied feedback and Jon also gets a chance to try his ideas.

There are several types of edge cases in the input to handle, and I feel that they are in scope now (or would eventually be).

jbeezley:
Did you intend to remove the composite keys for rows and columns or was that a mistake?

jbeezley:
I like the idea of making the column and row names computed properties.  We can probably do that as a smaller change independent of the rest.

I'm not ready to get rid of the pandas indices.  If we really need to do that, we should just use python csv parser.  This is how I envision it working.

1. On ingest read the raw table (`header=None, undex_col=None`)
2. Do some heuristic guessing of the row/column types
3. Pass the raw table and guesses to the user for modification
4. Reparse the tables using the now assumed correct row/column types

The last step could work as follows:
```python
def reparse_csv(table, rows, columns):
    return pandas.read_csv(
        BytesIO(table.iloc[rows, columns].to_csv(header=False, index=False).encode()),
        index_col=0
    )  
       

def read_tables(raw_table, index_row, index_column, data_rows, data_columns,
                metadata_rows, metadata_columns):
    data_table = reparse_csv(raw_table, [index_row] + data_rows, [index_column] + data_columns)
    measurement_metadata = reparse_csv(
        raw_table, [index_row] + metadata_rows, [index_column] + data_columns)
    sample_metadata = reparse_csv(
        raw_table, [index_row] + data_rows, [index_column] + metadata_columns)
    return data_table, measurement_metadata, sample_metadata
```

Using your example csv file, I get
```python
read_tables('index_test.csv', 2, 1, [4, 5, 6, 7, 8, 9, 11, 12], [3, 4, 5, 6, 7], [3], [2])                                                            

              1,6-Anhydro-beta-D-glucose  1-Methylnicotinamide  2-Aminobutyrate  2-Hydroxyisobutyrate  2-Oxoglutarate
Sample                                                                                                               
PIF_178                            40.85                 65.37            18.73                 26.05           71.52
PIF_087                            62.18                340.36            24.29                 41.68           67.36
PIF_090                           270.43                 64.72            12.18                 65.37           23.81
NETL_005_V1                       154.47                 52.98           172.43                 74.44         1199.91
PIF_115                            22.20                 73.70            15.64                 83.93           33.12
PIF_110                           212.72                 31.82            18.36                 80.64           47.94
NETCR_014_V1                       31.50                  6.82             4.18                 12.94           25.03
NETCR_014_V2                       51.42                 30.27             7.54                 34.81           80.64

       1,6-Anhydro-beta-D-glucose 1-Methylnicotinamide 2-Aminobutyrate 2-Hydroxyisobutyrate 2-Oxoglutarate
Sample                                                                                                    
NaN                         META1                 MET2           META3                META4          META5

                 Label
Sample                
PIF_178       cachexic
PIF_087       cachexic
PIF_090       cachexic
NETL_005_V1   cachexic
PIF_115       cachexic
PIF_110       cachexic
NETCR_014_V1  cachexic
NETCR_014_V2  cachexi
```
These tables all have the correct headers and sample names.  These become the normalized versions of the tables passed to the analysis/visualization functions.

subdavis:
Thanks for that example.  I think I understand what you're suggesting.

> 1. On ingest read the raw table (header=None, undex_col=None)

You've helped me realize that the case I'm making against Pandas is only about this step.  I don't think Pandas provides any value on ingest, so the normal csv parser may be better suited here.

> 2. Do some heuristic guessing of the row/column types

This step will be easier with the simple iterables that come out of python csv.  `for cell in table.iloc[:,col_half].fillna('').astype(str):` makes my head hurt.

> 3. Pass the raw table and guesses to the user for modification

:+1: 

> 4. Reparse the tables using the now assumed correct row/column types

So the input is 1) a raw CSV on disk and 2) the data models for CSVFile, TableRow, and TableColumn, and these 3 data tables are a derived product rather than a source of truth?  I'm totally in favor of this approach.  

As derived products, we could totally cache them on disk but would need some way to invalidate the cache if a user changes the Row/Column labels or primary keys.  Seems premature.  Are you just doing that Pandas parsing for every API query?

----------

> Did you intend to remove the composite keys for rows and columns or was that a mistake?

That was intended.  The composite key will error for collisions in the following cases:

* metadata columns have no headers in primary key row.  Example:

```text
,,"1,6-Anhydro-beta-D-glucose",1-Methylnicotinamide
PIF_178,cachexic,40.85,65.37
PIF_087,cachexic,62.18,340.36
```

* string duplicates in primary key row, because something in metadata differentiates them.  Example:

```
,,"1,6-Anhydro-beta-D-glucose","1,6-Anhydro-beta-D-glucose"
,Source,Spectrorama 9000,Thing in the basement from 1973
PIF_178,cachexic,40.85,65.37
PIF_087,cachexic,62.18,340.36
```

The first case definitely needs to be handled.  There's no reason in my mind why the second case shouldn't also work.  The fewer unnecessary errors we can show users, the better.  We could raise this as a warning on data validation on the cleanup page, but I believe the user should have the option to ignore that warning.

jbeezley:
> I don't think Pandas provides any value on ingest, so the normal csv parser may be better suited here.

I think it still does.  There is better handling of comments, empty lines, character encoding, missing columns/invalid rows.  Plus, using different parsers could result in different tables... e.g. raw csv parser creates an empty row and pandas automatically removes it.  Unless there is something specific that pandas *can't* handle, I think it's better to use it everywhere.

> This step will be easier with the simple iterables that come out of python csv. for cell in table.iloc[:,col_half].fillna('').astype(str): makes my head hurt.

I think you're just biased against pandas :).  The primary benefit is the read -> write loop generates a normalized version of a csv file that is consistent with what R expects.  Pandas was design to mimic R after all.

> As derived products, we could totally cache them on disk but would need some way to invalidate the cache if a user changes the Row/Column labels or primary keys. Seems premature. Are you just doing that Pandas parsing for every API query, then passing all 3 to transformations and visualizations?

I'm not worrying about performance until we have a better idea where the bottle necks actually are.  There are clearly repeated computations that we can optimize.  When we get there, I might try to use memcached.

> That was intended. The composite key will error for collisions in the following cases:

I don't see the problem with composite keys.  The unique constraint was removed when I made the the names derived.  The composite keys are `(csv_file_id, row/column_index)`.  The index in this case is just a 0-based integer describing the row/column number; that can't be duplicated.

subdavis:
> Pandas

Those are all fair points.

>  The unique constraint was removed when I made the the names derived.

I admittedly hadn't read that PR yet.  That takes care of my concerns.  Thanks.

With those issues settled, there's nothing really in this PR of value aside from a suggestion for 2: Do some heuristic guessing of the row/column types.  I just iterate through the middle row and middle column and label based on what data types are found there.

I'll leave the branch for reference and close this PR.

Thanks for the discussion, Jon!

jbeezley:
I still need to handle mutation of index rows/columns, but this work was helpful to clarify some of the issues.

jeffbaumes:
Agreed @jbeezley. Other benefits of using pandas parsing:
* It is probably the most resilient code path, because people worldwide are doing pandas csv reading lots more often than using the builtin Python csv parser.
* We ideally want to "test ingest" using the same libraries that will be doing the stats processing. We don't want the csv read to succeed during data cleanup then have it fail during stats processing (or have it prematurely fail if it would have succeeded in stats processing). We will probably want to continue to support pandas algorithms, and may even want to eventually do a "test ingest" in R to ensure downstream stats algorithms will work as expected.

PR #39, opened 2019-03-08
Add a d3 based PCA component with animated transitions

waxlamp:
@subdavis, @jbeezley: PTAL

If you upload a CSV and then continue past the cleanup view, you'll see a simple scatter plot component that should support the data transformations in the radio buttons.

Currently it only works with the normalization group at the top. I'm not sure if this is because I made a mistake or the transform functionality is not complete yet.

I'd love some feedback on my Vue practice (extending the Transform component with another view into VueX state; methods of triggering updates to the PCA component; etc.).

Also, I need some help adding a data ellipse to the PCA component: is that something that can be sent back from the pca endpoint along with the points themselves?


jbeezley:
Could you rebase your changes on master?  The `/plot/pca` endpoint should be there now.

> I'd love some feedback on my Vue practice (extending the Transform component with another view into VueX state; methods of triggering updates to the PCA component; etc.).

What you've done looks fine to me for now.  @subdavis can probably give better advice than me about interacting with the Vuex state.

> Also, I need some help adding a data ellipse to the PCA component: is that something that can be sent back from the pca endpoint along with the points themselves?

The computation is fairly simple (just eigenvectors of the covariance) and could be done client side, but it's probably easier to do it server side, bu we'll need an API change to return this along with the data.  What is the most convenient representation of an ellipse?  center + vector for major/minor axis?

waxlamp:
Really the best thing is probably to return the eigenvectors themselves (not all of them, I suppose, but the top `k` or something). Then the client can compute a data ellipse to display, but the data could also drive a loadings plot or other things as well.

I will take a look at `sklearn`'s PCA routine and see if I can extend this PR myself, and will ping you for help if I run into trouble.

jbeezley:
I'm going to start experimenting with doing this computation in R.  It may be best just to hard code some numbers for now.

subdavis:
Going to take a good look at this on the flight home.   I'm on West coast time, so feel free to not wait for me.

Thanks!

web/src/components/views/Transform.vue
```diff
@@ -55,17 +60,38 @@ export default {
     txTypeOrNull(tx) {
       if (tx && 'transform_type' in tx) return tx.transform_type;
       return null;
-    }
+    },
+    async loadPCAData (csv) {
+      const pcaData = await CSVService.getPlot(csv, 'pca');
+      this.points = pcaData.data;
+    },
+    async loadPCADataShift () {
```
subdavis: Where is this function called?

web/src/components/views/Transform.vue
```diff
@@ -55,17 +60,38 @@ export default {
     txTypeOrNull(tx) {
       if (tx && 'transform_type' in tx) return tx.transform_type;
       return null;
-    }
+    },
+    async loadPCAData (csv) {
+      const pcaData = await CSVService.getPlot(csv, 'pca');
```
subdavis: IMO, this load should go through Vuex.

I consider the PCA of <dataset_id> as part of global state, and vuex should handle what conditions cause PCA to get reloaded.  Putting the watched property on transformed below mixes state logic between the store and the component.  If we reuse PCA (which I'm fairly confident we will) then we have to duplicate the knowledge that "PCA state depends on transformed state".  If the relationship changes, we have to fix it in more than one place. 

In general, using REST services from within a Vue component throws flags for me (as is using watchers on vuex state).  I was sloppy with the image URL because we never intended to keep it and it would be removed as soon as PCA was ready.  I can remove that image in another PR.

After that long-winded digression though, it isn't really a big deal for this PR.  This code is easy to understand, and I think graph state management is a whole other concern that can be addressed in a separate PR because I have an idea how to do it and I'd like feedback from the group because my way may not be the best way.

web/src/components/views/Transform.vue
```diff
@@ -95,6 +121,9 @@ v-container(fill-height)
             v-radio(v-for="m in scaling_methods", :label="m.label",
                 :value="m.value", :key="`scale${m.value}`")
     v-layout(column).pa-2
+      v-card
+        vis-pca(:width="500" :height="400" :points="points")
```
subdavis: Should probably address window resize, probably in another PR.  Probably when we understand more about the layout.

subdavis:
Just got to set this up locally and play with it.  Its so cool!

subdavis:
I actually get a console error:

```
parse.js?da75:21 Error: <g> attribute transform: Expected number, "…8977614) rotate(NaN) scale(99.10…".
```
 
Originally though it was something to do with points being initially empty, but this is actually thrown after the update.  Do you see that error too?

web/src/components/views/Transform.vue
waxlamp: No, this was just a testing function I was using to simulate updates. Removed in e0b0e4c.

web/src/components/views/Transform.vue
waxlamp: Can you file an issue for this maybe? I am inclined to do it right in this PR, but if you have a more general fix that requires some discussion/education, maybe it's better in a separate PR.

web/src/components/views/Transform.vue
waxlamp: Good point. I opened #43 for this.

web/src/components/views/Transform.vue
waxlamp: Thanks @subdavis, I see #44 now.

PR #47, opened 2019-03-15
Vis pca linted

subdavis:
Hey there @waxlamp

I merged that linting PR, then took the liberty of rebasing yours onto master and adding the linting fixes here too.

I also removed that silly bar graph.  This should be exactly the same as your code, except with linting fixes.


web/src/components/views/Transform.vue
```diff
@@ -44,19 +44,24 @@ export default {
...
       norm(state) { return this.txTypeOrNull(state.datasets[this.dataset_id].normalization); },
       trans(state) { return this.txTypeOrNull(state.datasets[this.dataset_id].transformation); },
       scaling(state) { return this.txTypeOrNull(state.datasets[this.dataset_id].scaling); },
-      transformed(state) { return state.datasets[this.dataset_id].transformed },
+      transformed(state) { return state.datasets[this.dataset_id].transformed; },
     }),
     boxUrl() { return CSVService.getChartUrl(this.dataset_id, 'box'); },
     loadingsUrl() { return CSVService.getChartUrl(this.dataset_id, 'loadings'); },
   },
+  watch: {
```
waxlamp: Is this reordering enforced by a linter? More generally, is there style guidance on how to order the "sections" of a Vue component's JS code?

web/src/components/vis/VisPca.vue
```diff
@@ -1,10 +1,24 @@
...
-      <g class="plot"></g>
-      <g class="ellipse" transform="translate(0, 0) rotate(0) scale(1, 1)">
-        <circle cx="0" cy="0" r="1" style="fill: none; stroke: black;" vector-effect="non-scaling-stroke" />
+      <g class="axes" />
+      <g class="plot" />
+      <g
+        class="ellipse"
+        transform="translate(0, 0) rotate(0) scale(1, 1)"
+      >
+        <circle
```
waxlamp: I'm happy to adopt this style in writing templates, but is this actually more legible than the single line version it replaced?

web/src/components/vis/VisPca.vue
```diff
@@ -164,20 +186,20 @@ export default {
...
       const eigval = [
         trace / 2 + Math.sqrt(trace * trace / 4 - det),
-        trace / 2 - Math.sqrt(trace * trace / 4 - det)
+        trace / 2 - Math.sqrt(trace * trace / 4 - det),
       ];
 
-      const eigvec = Math.abs(xy) < 1e-10 ? [[1, 0], [0, 1]] :
-        [[eigval[0] - yy, xy],
-         [eigval[1] - yy, xy]];
+      const eigvec = Math.abs(xy) < 1e-10 ? [[1, 0], [0, 1]]
```
waxlamp: The colon on the next line is better; should we move the identity matrix expression to its own line too? Or, in general, are we better off dumping the `?:` operator here?

web/src/components/views/Transform.vue
```diff
@@ -115,6 +114,5 @@ v-container(fill-height)
                 :value="m.value", :key="`scale${m.value}`")
     v-layout.pa-2(column)
       v-card
-        img(:src="`${boxUrl}?cachebust=${norm}${trans}${scaling}`", style="width: 100%;")
```
waxlamp: Before we actually get rid of this, I'd like to have a quick discussion about the value of the plot itself. If it's needed in the app, we should talk about requirements for a better version; if it was simply a vis placeholder, then let's get rid of it. (A comment made by maybe Jeff made me think we *do* need a boxplot on this screen.)

waxlamp:
There's also one other concern: I don't think my eigenspace computations are correct. I will try to fix the immediate issue that's causing crashes in some circumstances, but generally speaking, my procedure almost always reports the identity matrix as the eigenvectors, and I'm pretty sure that's not correct, if only because we'd expect some variation in the eigenspaces of totally different test data.

This may become moot if we move to computing both the PCA results and the eigenvectors of the resulting data all using R, but that's something @jbeezley can comment on.

jeffbaumes:
I'd recommend we keep all stats math out of JS if possible and centralize it in Python and R. It will probably be difficult enough keeping things straight across those two languages.

waxlamp:
I agree, @jeffbaumes. I went ahead and did this just so I could mock out the ellipse, and fully intended to back the math out later.

web/src/components/views/Transform.vue
subdavis: Yeah, its enforced.  https://vuejs.org/v2/style-guide/#Component-instance-options-order-recommended -- priority C so not a huge deal.

web/src/components/vis/VisPca.vue
subdavis: I believe >100 chars was the reason for breaking it, not legibility.   Single line version is fine for <100 characters.

web/src/components/vis/VisPca.vue
```diff
@@ -80,16 +103,16 @@ export default {
       const xrange = minmax(points.map(d => d.x), 0.1);
       const yrange = minmax(points.map(d => d.y), 0.1);
 
-      const x = scaleLinear()
+      const scalex = scaleLinear()
```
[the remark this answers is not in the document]
subdavis: Shadowing.  x and y are shadowed on 177/178.

web/src/components/vis/VisPca.vue
subdavis: I'm a really bad offender of overusing ternary, so I'm probably not the one to ask.  I do like the look of `? / :` each on their own line personally, but have never checked official style

web/src/components/views/Transform.vue
subdavis: @jeffbaumes do we need a boxplot on this screen?   My potentially faulty memory is that I pulled the boxplot out of out of metaboanalyst, but that we were more focused on a PCA.

That box plot had badly labeled axis and looked terrible.  If we actually need one, I agree with @waxlamp  that we should open a new issue for one to do in D3.

web/src/components/vis/VisPca.vue
subdavis: Probably worth mentioning that the rest of the project uses pug, which is what I've configured the linter for.

I made that call without asking anyone (sorry), but we should probably be consistent for the future.  If you're not happy with pug as a template language, we can move everything back to html.  Zach and I (and maybe others?) like it a lot, but @matthewma prefers plain html, so it's not a unanimous thing.

web/src/components/views/Transform.vue
waxlamp: Thanks for pointing me to the style guide; will take a look.

web/src/components/vis/VisPca.vue
waxlamp: Ah I can redo this in Pug.

I'm a big fan of it actually, but when I did the Vue tutorials I also liked using HTML templates.

web/src/components/views/Transform.vue
jeffbaumes: Yeah, I'd say to get rid of the box plot. If it is requested, we should find a way to make a dynamic box plot client-side.

web/src/components/vis/VisPca.vue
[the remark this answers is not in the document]
waxlamp: In the arrow functions for `xMean` and `yMean`? I'm surprised the linter cared about that (since lexically, this isn't really a problem, unless those functions meant to reference the outer `x` and `y`). Stylistically, the `x` and `y` in those arrow functions ought to both be some other, more generic name, but I'm fine with changing the scale variable names to what you've got.

web/src/components/vis/VisPca.vue
waxlamp: Let's not overthink it and just leave this as it is now :cat:

waxlamp:
@jeffbaumes, would you prefer to back the clientside stats stuff out now, and replace the ellipse code later after we're computing it serverside?

subdavis:
Can't tell if this is CircleCI having a bad day or I've broken the build.  Seeing errors about the yarn cache after no changes to the package.json or yarn lock....

web/src/components/vis/VisPca.vue
waxlamp: I think we've got ourselves a rule conflict: the linter wants all attributes on a single line, but also wants to enforce a max line length. For my `circle` element, I have to break one rule or the other. I thought I'd be able to break the line between the attributes (due to the `multiLineSeparator` being specified) but it doesn't seem to be working. I'll push the thing that's breaking and maybe someone can take a look at it.

subdavis: Just needs to change the indent.  line overhang like that requires the overhang to be twice indented from the first line.

subdavis: See b57a7f9

web/src/components/views/Transform.vue
waxlamp: Thanks @jeffbaumes!

web/src/components/vis/VisPca.vue
waxlamp: Ah gotcha, thanks.

PR #48, opened 2019-03-15
State Cleanup, Edit Primary Keys

subdavis:
This PR includes

* minor fixes for vuex state
* ability to edit primary keys as discussed in another issue.
* ability to load state from hash route ID.

### DIFF
--- a/metabulo/models.py
+++ b/metabulo/models.py
@@ -333,13 +333,6 @@ class ModifyColumnSchema(Schema):
             raise ValidationError('JSON params required', data=data)
         return data
 
-    @post_load
-    def validate_type(self, data):
-        if data.get('column_type') == 'key':
-            raise ValidationError(
-                'Setting the key not yet supported', data='key', field_name='column_type')
-        return data
-
 
 class TableRow(db.Model):
     csv_file_id = db.Column(
@@ -374,14 +367,6 @@ class ModifyRowSchema(Schema):
     def validate_not_empty(self, data):
         if data is None or data == {}:
             raise ValidationError('JSON params required', data=data)
-        return data
-
-    @post_load
-    def validate_type(self, data):
-        if data.get('row_type') == TABLE_ROW_TYPES.INDEX:
-            raise ValidationError(
-                'Setting row primary key not yet supported',
-                data=data.get('row_type'), field_name='row_type')
         return data
 
 
--- a/metabulo/views.py
+++ b/metabulo/views.py
@@ -5,9 +5,9 @@ from flask import Blueprint, current_app, jsonify, request, Response, send_file
 from metabulo import opencpu
 from metabulo.models import CSVFile, CSVFileSchema, db, \
     ModifyColumnSchema, ModifyRowSchema, \
-    TableColumn, TableColumnSchema, TABLE_COLUMN_TYPES, \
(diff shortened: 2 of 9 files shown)


jbeezley:
I get an error when trying to build.
```
metabulo/web/src/components/CleanupTable.vue:9:11
    7|     thead
    8|       tr
  > 9|         th<!--empty-->
-----------------^
    10|         th.control(v-for="(col, idx) in dataset.column.labels")
    11|           select.pa-1(
    12|               :value="getDisplayValue('column', idx)",

Unexpected token `text-html` expected `text`, `interpolated-code`, `code`, `:`, `slash`, `newline` or `eos`
```

jbeezley:
One thing that is missing here is handling of tables that are missing a header row or index column.  For example, if you were to change the header row to "data", then an assertion error will be raised from [models.py#L105](https://github.com/girder/metabulo/blob/master/metabulo/models.py#L105).

I've been putting of consideration of how to handle this.  We could either add support somehow, or assert that if the header row is changed to data, then another row is changed to header in the same transaction.

subdavis:
I thought about that too, and my hunch was that validation should be a separate feature, and the cleanup UI won't let you move on until validation passes -- this would be most similar to simca.

I'm not sure that CSVFile model should require a header or key row/col.  If our guessing heuristic can't come up with what should be a header, It seems more helpful to show the user an invalid table with an error message that "maybe this table isn't valid for processing" and allow the user to either try and set a header or return to the upload screen.  

Basically the model should allow invalid/missing metadata and offer validation.  Might be nice to return this validation output for every metadata update so no extra request is needed.

In the future, we can do more advanced validation for cases where user set non-numeric row/columns as data and other stuff.

jeffbaumes:
Note: the below assumes we are good with moving away from the Simca interface style. If we/Tom thinks similarity to Simca is important, we should go more along the path of what @jbeezley and @subdavis are describing here. What is described below is a "don't let users do things that don't make sense" model as opposed to a "let users do whatever they want and signal errors" model.

Could we move to having more of this right on the csv model? Seems appropriate to do an expanded mockup here. If we go to something like this UI structure (note: this assumes major departure from the simca UI), we could make some simplifications to the model:

image

Dotted lines/icons show the extent of selection. For example, you can't make the data block include the id. The id works like a radio button control with single selection and can't go outside the metadata. You also can't exclude the id column. 

The main assumption is that all metadata should appear before any data (this seems intuitive but is something to confirm with Tom).

If we assume something like this UI, it seems the csv model could be extended to include id_column (int), id_row (int), data_start_column (int), data_start_row (int). Then you would have another table which is just the exclusion table for rows/columns. These exclusions are non-trivial, since you need to know whether the user marked it manually, or if an algorithm computed it (there will be two steps in the workflow to handle these cases).

subdavis:
Fixed the build error.

jbeezley:
Now that I play with this some more, I'm remembering that there are a number of places that break if primary keys change.  I should have a change today that fixes them.  For now, we can merge this if you want so I can use it as a base.

@jeffbaumes
That is a good summary of some of the discussions we had at the last meeting.  Could you add it as a new issue so we can discuss it?  This is just a small cleanup PR that I don't want to hold up.

PR #55, opened 2019-03-22
Always return the raw table through the rest api

### LINKED ISSUE #54: Server rearranges CSVFile
For the CSV "complex" example described @ https://github.com/girder/metabulo/pull/36#issuecomment-470980992 it the CSV is being reordered in a way that the client cannot reverse.

Regardless of how the server handles keys / parsing / etc. reordering should be hidden from the user.  I think reordering logic should probably be confined to the server.

#### Original CSV

Screenshot_20190321_212737

#### After page refresh (CSV is loaded from /csv/<id> endpoint)
Screenshot_20190321_212807

All the metadata are still in the same places, but the rows have moved around.

jbeezley:
The `table` property was really intended for internal use.  This renames properties `raw_table` to `table` and `table` to `indexed_table` on the csv_file model.  As a result, the table returned through the REST API for the cleanup UI will always be the same.

Fixes #54

### DIFF
--- a/metabulo/models.py
+++ b/metabulo/models.py
@@ -35,13 +35,13 @@ TABLE_COLUMN_TYPES = TableTypes('key', 'metadata', 'measurement', 'masked')
 TABLE_ROW_TYPES = TableTypes('header', 'metadata', 'sample', 'masked')
 
 
-def _guess_table_structure(raw_table):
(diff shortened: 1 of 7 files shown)


subdavis:
@jbeezley thanks!

PR #58, opened 2019-03-22
Draw PCA ellipses for data groups

waxlamp:
This PR depends on #57

### DIFF
--- a/web/src/components/vis/VisPca.vue
+++ b/web/src/components/vis/VisPca.vue
@@ -36,8 +36,8 @@ function covar(xs, ys) {
 }
 
 function unit(matrix) {
-  return matrix.map(row => {
-    const mag = Math.sqrt(row[0]*row[0] + row[1]*row[1]);
+  return matrix.map((row) => {
+    const mag = Math.sqrt(row[0] * row[0] + row[1] * row[1]);
     return row.map(d => d / mag);
   });
 }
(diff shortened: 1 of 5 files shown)


waxlamp:
[This bit](https://github.com/girder/metabulo/blob/6401f8ccd4e2f505139917ad69e81e6663c8f8d1/metabulo/views.py#L189-L194) does not feel very Pythonic to me. I'd like some feedback on how to express this better.

waxlamp:
@jbeezley, @subdavis: try this out and let me know what you think.

jbeezley:
I committed a change to simplify the label serialization.  If you could resolve the conflict, this LGTM.

web/src/components/vis/VisPca.vue
```diff
@@ -67,9 +70,14 @@ export default {
       const {
         width,
         height,
-        points,
+        rawPoints,
       } = this.$props;
 
+      const points = rawPoints.x.map(p => ({
```
subdavis: @waxlamp is `x` a list of points in 5 (or other specified) dimensions?  Why bother computing more than 2 or 3?  Also, why name that list `x`?

subdavis:
Looks great, only one question from me.

web/src/components/vis/VisPca.vue
waxlamp: I think it's named `x` here because that's what `sklearn` calls it when the PCA is performed (but @jbeezley can confirm that).

As for why 5, that's the default setting for the `pca` endpoint, and I was too lazy to change it to 2 in my call. It should possibly be 2 by default but that depends on the typical use case for PCA (which may not be simply visualizing the first two PCs).

waxlamp:
@jbeezley conflicts resolved, but it looks like I'm not an "authorized user" so I can't merge it myself :crying_cat_face:

PR #63, opened 2019-03-26
Add keys to v-for declarations

waxlamp:
@subdavis please take a look and let me know if the key I supplied for these `v-for` declarations are truly unique.


subdavis:
@waxlamp yeah they are.

I neglected to add the keys here because those lists are static, but I should have.  Thanks!

PR #64, opened 2019-03-26
Add tooltips for data points in PCA component

waxlamp:
This can be augmented or improved in various ways (in future PRs). Some suggestions:
- improve the general styling of the chart itself (I've already spoken a bit to @jtomeck about this)
- display more data in the tooltip
- let the tooltip expand to the natural size to display the requested data

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -12,6 +12,7 @@
     "@girder/components": "https://github.com/girder/girder_web_components#master",
     "axios": "^0.18.0",
     "d3-axis": "^1.0.12",
+    "d3-format": "^1.3.2",
     "d3-scale": "^2.2.2",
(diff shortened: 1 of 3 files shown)


waxlamp:
Note: I force pushed just now, in case anyone had this checked out.

subdavis:
I think it looks great -- I can merge whenever you're ready.

It might be nice to have bigger "hitboxes" for the tooltip activation, or maybe even bigger points?  Could make the points considerably larger if they were semi-opaque.  Not important for this PR though.

waxlamp:
I had a similar thought about the size of the points. I'll open an issue
for styling in general, and let's go ahead and merge this one now.

Thanks!

PR #66, opened 2019-03-27
Create a fixed ingestion pipeline

jbeezley:
This replaces the old dynamic transform pipeline.  For the moment, the
pipeline only contains the normalization step, but addtional steps can
be added by creating a new column on the csv_file table and including
a new step in the "CSVFile.apply_transforms" method.

The new API is

```
PUT /csv/:id/normalization json="minmax"
PUT /csv/:id/normalization json=null
```


metabulo/views.py
```diff
@@ -96,6 +97,22 @@ def delete_csv_file(csv_id):
 
     return '', 204
 
+# Ingest transforms
+@csv_bp.route('/csv/<uuid:csv_id>/normalization', methods=['PUT'])
+def set_normalization_method(csv_id):
+    csv_file = CSVFile.query.get_or_404(csv_id)
+    args = request.json
+    if args is not None and args not in NORMALIZATION_METHODS:
```
subdavis: Could we change to key-value pair, like `{ type: "minmax" }` or something?  Axios tries to get clever and sends mime type based on data payload, which results in `minmax` being sent as formdata.

subdavis:
Changed API to:

### PUT /csv/:id/normalization

```json
{
    "method": "string" | null
}
```

jbeezley:
Sounds good.  Go ahead and merge if the client is ready.

PR #70, opened 2019-03-28
Fix/graphstate

subdavis:
This PR reduces the amount of work a plot container must do to keep its plot up-to-date, and moves the logic for cached/stale points into vuex.

A component wanting to show a plot will need to have the following for each plot it wants to show.

```javascript
export default {
  computed: {
    pcaData() { return this.$store.getters.plotData(this.dataset_id, 'pca'); },
    pcaValid() { return this.$store.getters.plotValid(this.dataset_id, 'pca'); },
  },
  watch: {
    pcaValid: {
      immediate: true,
      handler(valid) {
        if (valid === false) {
          this.$store.dispatch(LOAD_PLOT, { dataset_id: this.dataset_id, name: 'pca' });
        }
      },
    },
  },
};
```

This is nice because:
* No component has to care which mutations cause the plot to become stale.
* Mutations from other components or parts of the app tree will still cause the proper refresh.
* Plots are only updated when they need to be, not necessarily between route changes or remounts.

### DIFF
--- a/web/src/components/vis/VisPca.vue
+++ b/web/src/components/vis/VisPca.vue
@@ -78,13 +78,8 @@ export default {
      */
     rawPoints: {
       required: true,
-      validator: (prop) => {
-        if (!prop) {
-          return true;
-        }
-        return typeof prop === 'object'
-            && ['x', 'labels'].every(key => key in prop);
-      },
+      validator: prop => !prop
+          || (typeof prop === 'object' && ['x', 'labels'].every(key => key in prop)),
     },
   },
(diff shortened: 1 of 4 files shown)


waxlamp:
Thanks @subdavis; will take a look later today.

subdavis:
I prematurely merged this.  Looks like the rebase ate some changes from Transform.vue

I'll open a follow-up to add them back.

waxlamp:
Oops. Let me know if I can help.

PR #73, opened 2019-03-28
Add a "group" column type

jbeezley:
@subdavis @jtomeck I added the new type to the cleanup table UI, but it could use some way to distinguish "group" columns from other metadata columns.


jbeezley:
Also, the new behavior for the import is to set the second column as the "group" column.

PR #74, opened 2019-03-28
Add table validation for better error handling

jbeezley:
This is an attempt at providing a standard way to communicate the table is not yet ready for transforms.  It adds:
* A `table_validation` property to the `CSVFile` model.  It is a list of problems or `None` if everything is okay.
* A standalone endpoint to get the validation property.

I added the endpoint for a quicker way to get this information after posting row/column type changes.  Alternatively, we could find a way to pass this information in the response to those endpoints.  @subdavis Do you think the new endpoint is sufficient or would it be better to provide this information with all endpoints that are capable of modifying the validation state?

### DIFF
--- a/metabulo/models.py
+++ b/metabulo/models.py
@@ -64,6 +64,19 @@ class CSVFile(db.Model):
     meta = db.Column(JSONType, nullable=False)
 
     @property
+    def table_validation(self):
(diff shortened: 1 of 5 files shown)


jbeezley:
I'll merge this for now as I need some of the error handling for other things.  We can revisit the endpoints if necessary.

PR #79, opened 2019-03-28
Use dogpile for caching CSV tables

jbeezley:
This replaces the old method caching (private attributes) with dogpile.
If pylibmc is installed it will use that, otherwise it falls back to an
in-memory cache for development and testing.  This will become more
important when we start adding more caching to the various processing
stages.

I think the deployment changes will work, but maybe not.  Hold my beer.


mgrauer:
Make sure to take your beer back.

Issue #84, opened 2019-04-01
Dataset Validation Schema


subdavis:
To produce the views in #83, some kind of schema for data validation will be necessary.

It would be ideal to shape it something like this:

```javascript
      validation: [
        { severity: 'error', type: 'primary-key', title: 'Primary key missing' },
        { severity: 'error', type: 'group-tag', title: 'Group tag missing' },
        {
          severity: 'error', type: 'missing-data', title: 'Missing data', data: [3, 4, 15, 18, 22, 23, 24, 25, 26, 27, 28, 29],
        },
        {
          severity: 'warning', type: 'low-variance', title: 'Low variance', data: [5, 6, 7, 9, 12, 14],
        },
      ],
```

Where `severity`, `type`, and `title` are required and `data` could either be a list of the column indexes where that applies or a list of objects with more data about each instance, such as percent missing or r^2, respectively.

```javascript
data: [
  { index: 34, missing: .34 },
]
```

We could also encode a map of `type -> (title, description, severity)` in the client code and remove those as requirements from the schema, but that would require client changes if the enum of possible validation errors changed.

cc @jbeezley is this along the lines of what you were thinking?

Side note: right now  there's a table validation error `Error applying transforms`.  I don't think this should be a validation error.  If transforms fail and the server can't determine a validation reason, that should probably be some kind of 400 or even 500 error.

jbeezley:
I can work with that schema.

RE: missing data.  It will be a warning because the imputation step will automatically fill in all missing values.  Also, the R script keeps track of missing data per group.  So there could be multiple missing data warnings per column that each apply to a subset of rows.  

> We could also encode a map of type -> (title, description, severity) in the client code and remove those as requirements from the schema, but that would require client changes if the enum of possible validation errors changed.

I'm okay with keeping it server-side, but I doubt we will be able to decouple the definition from the client side implementation.  The client will need to understand the details of the error to suggest solutions or provide more context.

> Side note: right now there's a table validation error Error applying transforms. I don't think this should be a validation error. If transforms fail and the server can't determine a validation reason, that should probably be some kind of 400 or even 500 error.

We could return an HTTP error code, but I'm imagining a hypothetical scenario where two parameters need to change at once in different steps.  (Setting each individually would cause an error in the pipeline.)  We would need to provide a single endpoint to set both parameters at once.

[...]

PR #85, opened 2019-04-02
Compute row and column statistics

jbeezley:
This adds `missing_percent` and `data_variance` properties on each row and column.  These get returned with the endpoints in case we want to use them in the UI.

I'm marking this as WIP for now because it is far too slow to get merged (adds about 3 seconds to each affected endpoint).  I have some ideas about smarter caching strategies that should help.  If not, I'll have to do some refactoring to reduce the number of duplicate computations occurring.

### DIFF
--- a/metabulo/app.py
+++ b/metabulo/app.py
@@ -5,7 +5,7 @@ from flask import current_app, Flask, jsonify
 from marshmallow import ValidationError
 from werkzeug.middleware.proxy_fix import ProxyFix
 
-from metabulo.cache import region
+from metabulo.cache import clear_cache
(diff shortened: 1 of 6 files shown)


jbeezley:
TIL: Dogpile doesn't include the "self" argument in its computed key for class methods.  I had to work around this to get caching working correctly.

The new caching speeds up the requests to a reasonable ~500ms.  The cache is aggressively purged and only local to each request, so there is still room for speed ups if it becomes necessary.

The code included here will not necessarily be useful to the client, but it includes infrastructure that will help in resolving #84.

PR #86, opened 2019-04-02
Radio buttons

subdavis:
Try with radio buttons.


subdavis:
Screenshot_20190402_130351

PR #88, opened 2019-04-03
Implement validation API for fatal errors

jbeezley:
This implements the API described in #84 with a few minor changes.  The response contains:
* `title`: Some short user-friendly description.  We may want to move this to the client.
* `type`
* `severity`
* `context`: `table`, `row`, or `column` (maybe `group` and `element` in the future)
* `row_index`, `column_index`: If applicable from the context
* `data`: Error specific data (e.g. the missing fraction for a missing data warning)

@subdavis Most of the planned errors and warnings are not yet reported, but I wanted to get this in so you can work on wiring it and let me know if there are any problems.  Once more of the error detection is implemented, I'll use it to improve the default table parsing as well.


subdavis:
:+1:

PR #93, opened 2019-04-04
Adds Batching to set axis keys

### LINKED ISSUE #91: Batch Modify Row/Column
It would be nice to have an endpoint 

`PUT /csv/<uuid:csv_id>/batch/label`

with body:

```jjavascript
[
  { context: 'row', index: 0, label: 'metadata' },
  { context: 'column', index: 4, label: 'key'},
]
```

I actually don't see the utility of the regular endpoints if we had this -- it would be easier to use batch for everything.  It would also be nice to not have to encode type info into the key names: `row_type` vs `type`.

I'm going to take a stab at this.

subdavis:
There are a lot of tests to fix, so I'll do that as soon as we confirm that the new API is OK.

Drops about 100 LoC, so that's cool

resolves #91

### DIFF
--- a/metabulo/views.py
+++ b/metabulo/views.py
@@ -5,8 +5,8 @@ from marshmallow import ValidationError
 import pandas
 
 from metabulo import opencpu
-from metabulo.models import CSVFile, CSVFileSchema, db, \
-    ModifyColumnSchema, ModifyRowSchema, \
+from metabulo.models import AXIS_NAME_TYPES, CSVFile, CSVFileSchema, db, \
+    ModifyLabelListSchema, \
     TABLE_COLUMN_TYPES, TABLE_ROW_TYPES, \
     TableColumn, TableColumnSchema, TableRow, \
     TableRowSchema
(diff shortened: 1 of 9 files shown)


metabulo/views.py
```diff
@@ -141,22 +140,34 @@ def get_column(csv_id, column_index):
...
+                csv_file.key_column_index = index
+            elif label == TABLE_COLUMN_TYPES.GROUP:
+                csv_file.group_column_index = index
+            setattr(column, 'column_type', label)
+            db.session.add(column)
+
     db.session.add(csv_file)
     db.session.commit()
-    return jsonify(table_column_schema.dump(column))
+    return jsonify(csv_file_schema.dump(csv_file))
```
jbeezley: I would catch exceptions in this endpoint and call `db.session.rollback()` after any error.

tests/test_row_column.py
```diff
@@ -137,135 +135,127 @@ def test_modify_column(client, table):
...
+#     assert resp.status_code == 200
+#     assert (CSVFile.query.get(csv_id).measurement_table.columns == ['header2', 'header3']).all()
+
+#     resp = client.put(
+#         url_for('csv.batch_modify_label', csv_id=csv_id),
+#         json={'changes': [{'context': 'column', 'index': 0, 'label': 'measurement'}]}
+#     )
+#     assert resp.status_code == 200
+#     assert (CSVFile.query.get(csv_id).measurement_table.columns ==
+#             ['header1', 'header2', 'header3']).all()
```
jbeezley: It's okay to get rid of most of this.  There is fairly good coverage of the row/column related code at the model layer in `test_csv_file.py`.  I would replace it with one or two examples of non-trivial batch mutations.

subdavis:
@jbeezley rollback, new tests added.

PR #94, opened 2019-04-04
Change group validation error message

jbeezley:
We could allow empty group names, but it would require handling elsewhere in the code.  I'll delay doing so until it is requested.


subdavis:
LGTM.

PR #96, opened 2019-04-04
Fix missing key / fix an edge case

waxlamp:
attn @subdavis


subdavis:
Nice.  I keep forgetting keys lately.

PR #101, opened 2019-04-05
Use memcached for slow functions

jbeezley:
This reduces the response time for pca plots on the client by speeding up `GET /csv/:id`, `GET /csv/:id/plot/pca`, and `PUT /csv/:id/normalizaton` by ~10x--making the update when toggling normalization take about 100ms.

### DIFF
--- a/metabulo/views.py
+++ b/metabulo/views.py
@@ -5,6 +5,7 @@ from marshmallow import ValidationError
 import pandas
 
 from metabulo import opencpu
+from metabulo.cache import csv_file_cache
 from metabulo.models import AXIS_NAME_TYPES, CSVFile, CSVFileSchema, db, \
     ModifyLabelListSchema, \
     TABLE_COLUMN_TYPES, TABLE_ROW_TYPES, \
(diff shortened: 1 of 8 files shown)


metabulo/views.py
```diff
@@ -118,7 +125,7 @@ def set_normalization_method(csv_id):
         csv_file.normalization = method
         db.session.add(csv_file)
         db.session.commit()
-        return jsonify(csv_file_schema.dump(csv_file))
+        return jsonify(method)
```
subdavis: @jbeezley @waxlamp  could we revert this change?

I'm not aware of any scenario from the UI where this speeds things up.  Any time the norm/trans/scaling are changed, the UI will need to reload the CSV to refresh the graphs.  There are no scenarios where this endpoint will be hit without the other.

Breaking this into 2 API requests just adds extra latency.

waxlamp: Some context: I need to use the transformed values for the loadings plot, and was getting the name of the last applied transform/normalization/scaling operator in the `dataset.transformed` field, rather than the transformed data itself. As things stand now (after this change was made), `dataset.transformed` carries just about zero information.

jbeezley: It's fine with me, I'm too far removed to remember why I did it in the first place.

waxlamp: Cool. Rather than literally revert this change, I'll do it as part of my current branch.

Thanks @subdavis and @jbeezley!

PR #112, opened 2019-04-11
Add interactivity and hooks for warning validation

subdavis:


jbeezley:
This is just missing the `non-numeric-data` warning.  The default behavior is to coerce these elements to NaN.  It is currently a table-wide warning, but I could make it column, row, or even element specific if it makes more sense in the UI.

subdavis:
There are actually a couple of issues here I'll follow up with.

PR #115, opened 2019-04-11
Add integration with sentry

subdavis:
Looks like they have a new library to replace Raven with a simpler API surface.  :tada:

### DIFF
--- a/web/src/main.js
+++ b/web/src/main.js
@@ -2,6 +2,8 @@ import '@mdi/font/css/materialdesignicons.min.css';
 
 import Vue from 'vue';
 import Vuetify from 'vuetify';
+import * as Sentry from '@sentry/browser';
+import { Vue as SentryVue } from '@sentry/integrations';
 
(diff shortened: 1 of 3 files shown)


web/src/main.js
```diff
@@ -15,6 +17,11 @@ Vue.config.productionTip = false;
 
 ApiService.init();
 
+SentryInit({
+  dsn: 'https://9c7281f0dfc5402a953698256bd067e0@sentry.io/1437129',
+  integrations: [new SentryVue({ Vue })],
+});
```
jbeezley: Might want to only do this in the production environment.

subdavis: I believe the library handles this for you.... let me check.

jbeezley: I got a notification, but maybe you were testing in production mode.

subdavis: Nope, you're right.  Need to check myself.

PR #117, opened 2019-04-11
Only install sentry in production

subdavis:
@jbeezley Is this how you'd do it?


jbeezley:
Yeah, I think that should work.

PR #118, opened 2019-04-11
Fix problems detecting errors in header rows

jbeezley:
This makes the validation more picky about rows/columns used as primary keys.  Before this change, it could deal with duplicates by appending a unique value.  I found this breaks a number of internal assumptions that I *could* fix, but it is easier to just fail validation for now.


jbeezley:
BTW, I found this because of a sentry alert.  :tada:

PR #125, opened 2019-06-20
Add PCA loadings plot

waxlamp:
This PR adds a companion loadings plot to go with the PCA scores plot.

Some work still needs to be done:
- [x] refactoring of common code from the client code for the two plots, and the server code that drives them
- [ ] layout correctness/optimization: @jtomeck and @subdavis, could you see if the [templates for the plots](https://github.com/girder/metabulo/blob/loadings-plot/web/src/components/Transform.vue#L128-L136) look right?
- [ ] PCA validation: I'd like to sanity check whether these plots are actually computing the correct thing; @jbeezley, I could use your help with this. I will post a test case in the comments when I have one prepared.
- [ ] refactor the tooltip functionality into a mixin?


waxlamp:
@jbeezley (and whoever else has insight into this): here's a [CSV file](https://gist.github.com/waxlamp/658b63e925d50a4edb5c4d59ba2c514d) we can use to try to recreate this [PCA score and loadings plots](https://stats.stackexchange.com/questions/7860/visualizing-a-million-pca-edition/7862#7862). If you upload this file to Metabulo (using this branch, of course), and set the scaling to "Autoscaling", you'll see what looks eerily like a 180-degree rotated version of the plot in that CrossValidated post.

I'd like some help to figure out if this is correct, or if I might have some kind of off-by-π error in my code.

subdavis:
Looks awesome!   Going to check out #126 then I'll read this more thoroughly.

PR #126, opened 2019-06-21
Column styles

jtomeck:
Change alternating column colors to borders.

### DIFF
--- a/web/src/components/Cleanup.vue
+++ b/web/src/components/Cleanup.vue
@@ -276,6 +276,14 @@ v-layout.cleanup-wrapper(row)
         border-bottom: 2px solid var(--v-secondary-darken3);
       }
 
+      &.active {
+        &.metadata {
+          td {
+            box-shadow: inset 0 0 0 1px rgba(161, 213, 255, 0.15) !important;
+          }
+        }
+      }
+
       &.active td,
       td.active {
         background: linear-gradient(0deg,rgba(161, 213, 255, 0.2),rgba(161, 213, 255, 0.2));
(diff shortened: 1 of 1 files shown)


subdavis:
#127 adds some comments.

web/src/components/Cleanup.vue
```diff
@@ -286,75 +276,81 @@ v-layout.cleanup-wrapper(row)
...
         td {
-          background-color:var(--v-accent-lighten1);
+          background-color: var(--v-accent-lighten1);
+          box-shadow: inset 0 0 0 .5px var(--v-accent-base);
+
+          &.active {
+            color: white;
+
+            &.first{
+              border-left: 2px solid var(--v-secondary-darken3) !important;
```
subdavis: I can't tell that this important is necessary.

web/src/components/Cleanup.vue
```diff
@@ -286,75 +276,81 @@ v-layout.cleanup-wrapper(row)
...
+            }
+            &.last {
+              border-right: 2px solid var(--v-secondary-darken3) !important;
```
subdavis: Same with this one.

web/src/components/Cleanup.vue
```diff
@@ -185,9 +185,10 @@ v-layout.cleanup-wrapper(row)
 
 <style lang="scss">
 @mixin masked() {
-  background-color: var(--v-secondary-lighten1);
+  background-color: var(--v-secondary-lighten3);
+  box-shadow: inset 0 0 0 .5px var(--v-secondary-lighten2);
   font-weight: 300;
-  color: var(--v-secondary-darken1);
+  color: var(--v-secondary-base) !important;
```
subdavis: I think this one can be replaced if `tr.datarow, tr.datarow.header` is added at the very end.

subdavis:
Other than those 3 lines, this looks awesome and is way less complex than it was before, which I'm all in favor of.

subdavis:
Cool looks good to me!  Thanks.

PR #127, opened 2019-06-21
Remove important, change things to compensate

subdavis:
@jtomeck  Could we remove the usages of `!important`?  I think those should be avoided when possible.

Also, what do those `box-shadow` rules do?  I don't see any change with our without those rules.

Could you before/after maybe?  Ty.

### DIFF
--- a/web/src/components/Cleanup.vue
+++ b/web/src/components/Cleanup.vue
@@ -188,7 +188,7 @@ v-layout.cleanup-wrapper(row)
   background-color: var(--v-secondary-lighten3);
   box-shadow: inset 0 0 0 .5px var(--v-secondary-lighten2);
   font-weight: 300;
-  color: var(--v-secondary-base) !important;
+  color: var(--v-secondary-base);
 }
 
 .cleanup-wrapper {
@@ -196,7 +196,7 @@ v-layout.cleanup-wrapper(row)
 
   .radio-toolbar {
     .v-toolbar__content {
-      height: inherit !important;
+      height: inherit;
       padding: 0px;
     }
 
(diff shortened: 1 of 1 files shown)


web/src/components/Cleanup.vue
```diff
@@ -349,7 +334,8 @@ v-layout.cleanup-wrapper(row)
       }
     }
 
-    tr.datarow {
+    tr.datarow,
+    tr.datarow.header {
```
subdavis: As far as I could tell, none of the importants were necessary except for line 191, which this fixes.

jtomeck: @subdavis the `!important` seemed to be completely necessary, but I'll take another look.

In terms of the `box-shadow` properties, that is what I'm using to create the borders. There are multiple solutions to adding borders to tables:

1. Just add borders to the th/td's - This limits you to a minimum of a 2px border since each cell's 1px border would be adjacent to another cell's 1px border.
2. Figure out what side of each table element needs a border to avoid borders touching - For example, `table` could have `border-top` and `border-left` and each `tr` and `td` could have `border-bottom` and `border-right`. This gets complex when using column colors and adding borders when columns are selected.
3. `box-shadow` - In practice this is doing exactly what no. 1 of this list is talking about, however, `box-shadow` allows for .5px values, whereas `border` does not.

The third option seemed the best & easiest, but let me know if you'd prefer I explore the second option instead. In the meantime I'll look into the need for `!important` and try to get rid of some/all of them.

subdavis: Sorry, I don't think I was very clear.  

I was only asking about the 2 instances of box-shadow I removed.  https://github.com/girder/metabulo/pull/127/files has some changes, and as far as I can tell, after those changes are applied, nothing changes between your PR and mine.

PR #128, opened 2019-06-24
fixes for loadings-plot

subdavis:
@waxlamp first, forgive my meddling.  I did this because I wanted to get into the weeds a bit with your code and understand it, and this was the best way for me to do that.

If you'd like to simply read over it then make whatever changes you decide to keep by hand on your own branch, that won't bother me at all.

This branch consists of a refactor of the axisPlot mixin and a small fix for VisPca because the axes didn't update when scaling changed, so they'd get initialized once then not update.

There's also a container overflow fix (which only helps on Chrome, flexbox heights work slightly different on Firefox and I can't figure out the search terms to find any discussion of the problem).

## axisPlot bits.

First, I really liked the mixin here -- I think it was a good place to use one.  

The main issues I wanted to address were:
* that the mixin was creating properties on the instance in a non-reactive way and from within a method where it isn't obvious that the mixin owns those properties.
* that the properties it was creating were deterministic of the props, and could be factored into computed props.
* that the mixin was creating data properties that the parent would then use, the timings of which were hard to understand.  (e.g. I have to call `axisPlot` and then my component will have some new properties like `svg`.

I'm still learning mixins, but so far I've had a good experience with treating them like abstract base classes (minus the abstract methods bits).  I try to make my mixins attachable to any component:  they don't rely on any data or methods to be present on the components they modify.   In this case, a mixin can't use `$refs.svg` because mixins can't change/own templates.

PTAL and let me know what you think!

### DIFF
--- a/web/src/components/vis/VisPca.vue
+++ b/web/src/components/vis/VisPca.vue
@@ -165,7 +165,10 @@ export default {
         rawPoints,
       } = this.$props;
 
-      const { xyPoints } = this;
+      const {
+        xyPoints,
+        group,
+      } = this;
 
       // Set the axis labels.
       //
@@ -182,19 +185,18 @@ export default {
       //
       // Set up a colormap and select an arbitrary label to color the nodes.
       const cmap = scaleOrdinal(schemeCategory10);
-      const group = this.group;
 
       // Plot the points in the scatter plot.
       const tooltip = select(this.$refs.tooltip);
       const coordFormat = format('.2f');
-      const sel = svg.select('g.plot')
+      svg.select('g.plot')
         .selectAll('circle')
         .data(xyPoints)
         .join(enter => enter.append('circle')
           .attr('cx', this.scaleX(0))
           .attr('cy', this.scaleY(0))
           .attr('r', 0)
-          .style('fill', (d, i) => group ? cmap(rawPoints.labels[group][i]) : null)
+          .style('fill', (d, i) => (group ? cmap(rawPoints.labels[group][i]) : null))
           .on('mouseover', function mouseover(d) {
             select(this)
               .transition()
@@ -288,7 +290,7 @@ export default {
         .enter()
         .append('g')
         .classed('ellipse', true)
-        .attr('transform', (d) => `translate(${this.scaleX(d.xMean)}, ${this.scaleY(d.yMean)}) rotate(0) scale(1, 1)`)
+        .attr('transform', d => `translate(${this.scaleX(d.xMean)}, ${this.scaleY(d.yMean)}) rotate(0) scale(1, 1)`)
         .append('circle')
         .attr('cx', 0)
         .attr('cy', 0)
--- a/web/src/components/vis/VisPca.vue
+++ b/web/src/components/vis/VisPca.vue
@@ -98,6 +98,21 @@ export default {
         && Array.isArray(prop.source.columns),
     },
   },
+  data() {
+    return {
+      margin: {
+        top: 20,
+        right: 0,
+        bottom: 50,
+        left: 50,
+      },
+      xrange: [-1, 1],
+      yrange: [-1, 1],
+      xlabel: 'PC1',
+      ylabel: 'PC2',
+      duration: 500,
+    };
+  },
   computed: {
     xyPoints() {
       if (this.rawPoints) {
@@ -121,19 +136,9 @@ export default {
       if (newval) {
         if (!this.axisPlotInitialized) {
           const { xyPoints } = this;
-          this.axisPlot({
-            margin: {
-              top: 20,
-              right: 0,
-              bottom: 50,
-              left: 50,
-            },
-            xrange: minmax(xyPoints.map(p => p.x), 0.1),
-            yrange: minmax(xyPoints.map(p => p.y), 0.1),
-            xlabel: 'PC1',
-            ylabel: 'PC2',
-            duration: 500,
-          });
+          this.setRanges(xyPoints);
+          const svg = select(this.$refs.svg);
+          this.axisPlot(svg);
         }
 
         this.update();
@@ -141,26 +146,19 @@ export default {
     },
   },
   mounted() {
-    if (this.xyPoints) {
-      const { xyPoints } = this;
-      this.axisPlot({
-        margin: {
-          top: 20,
-          right: 0,
-          bottom: 50,
-          left: 50,
-        },
-        xrange: minmax(xyPoints.map(p => p.x), 0.1),
-        yrange: minmax(xyPoints.map(p => p.y), 0.1),
-        xlabel: 'PC1',
-        ylabel: 'PC2',
-        duration: 500,
-      });
-
+    const { xyPoints } = this;
+    if (xyPoints) {
+      this.setRanges(xyPoints);
+      const svg = select(this.$refs.svg);
+      this.axisPlot(svg);
       this.update();
     }
   },
   methods: {
+    setRanges(xyPoints) {
+      this.xrange = minmax(xyPoints.map(p => p.x), 0.1);
+      this.yrange = minmax(xyPoints.map(p => p.y), 0.1);
+    },
     update() {
       // Grab the input props.
       const {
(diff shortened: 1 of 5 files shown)


web/src/components/vis/mixins/axisPlot.js
```diff
@@ -11,46 +10,72 @@ function labelAxis(label, msg, xFunc, yFunc, rot) {
...
+  },
+  data() {
+    return {
+      axisPlotInitialized: false,
+      margin: {
+        top: 20,
+        right: 0,
+        bottom: 50,
+        left: 50,
+      },
```
waxlamp: `margin` is passed into `axisPlot()` below as a required statement from the vis component, so perhaps this initializer should contain `null`s instead of these numbers?

waxlamp: (possibly the same question for `xrange` and `yrange` below, though it makes sense to supply reasonable defaults for `xlabel`, `ylabel`, and `duration` IMO--even `xrange` and `yrange` might have reasonable defaults, but likely not `[-1.2, 1.2]`)

web/src/components/vis/mixins/axisPlot.js
```diff
@@ -86,16 +109,15 @@ export const axisPlot = {
...
-      this.setYLabel(ylabel);
+      this.setXLabel(svg, xlabel);
+      this.setYLabel(svg, ylabel);
 
       // Conclude the initialization process.
       this.axisPlotInitialized = true;
     },
 
-    setXLabel(msg) {
+    setXLabel(svg, msg) {
```
waxlamp: Would it be bad to "capture" `svg` during `axisPlot` as local data, then use it here, instead of forcing the caller to supply it? This is coming from my obvious background having grown up as an OOP programmer and wanting to encapsulate this bit of data, since it is not supposed to change over the lifetime of this mixin's instantiation.

web/src/components/vis/VisPca.vue
```diff
@@ -174,15 +172,16 @@ export default {
       // Compute the total variance in all the PCs.
       const totVariance = rawPoints.sdev.reduce((acc, x) => acc + x, 0);
       const pctFormat = format('.2%');
-      this.setXLabel(`PC1 (${pctFormat(rawPoints.sdev[0] / totVariance)})`);
-      this.setYLabel(`PC2 (${pctFormat(rawPoints.sdev[1] / totVariance)})`);
```
waxlamp: These two `set[X|Y]Label()` calls are needed to update the plot's labels. If we just place `svg` as the first argument to both, will this work as originally intended?

web/src/components/vis/mixins/axisPlot.js
```diff
@@ -11,46 +10,72 @@ function labelAxis(label, msg, xFunc, yFunc, rot) {
...
+  },
+  computed: {
+    dwidth() {
+      const { width, margin } = this;
+      return width - margin.left - margin.right;
+    },
+    dheight() {
+      const { height, margin } = this;
+      return height - margin.top - margin.bottom;
+    },
```
waxlamp: These two computed properties are FANTASTIC. I am mad that I didn't think to do this (still getting my head around Vue, obviously :-D).

web/src/components/vis/mixins/axisPlot.js
```diff
@@ -11,46 +10,72 @@ function labelAxis(label, msg, xFunc, yFunc, rot) {
...
-      // Create X and Y axis objects.
-      const axisX = axisBottom(scaleX);
-      this.axisX = axisX;
+    },
+    axisX() {
+      return axisBottom(this.scaleX);
+    },
+    axisY() {
+      return axisLeft(this.scaleY);
+    },
```
waxlamp: So, I do have an itch about `axisX()` and `axisY()` as updated (and, to a lesser extent, `scaleX()` and `scaleY()` above as well): these are D3 functions that return what is intended to be a persistent object for use in the rest of the code. In particular, they are *mutable* objects, which means that users of this mixin who are accustomed to the D3 way of doing things might mutate them in various ways, then find that the mutation semantics are weirdly ephemeral (or, perhaps the way the mixin mechanism works means that users of the mixin do not have access to these values; it's not super clear to me). However, this problem is mostly academic for now, because *my* current usage of these values doesn't rely on mutating them.

The other problem is of performance: it feels weird to keep reconstructing these as needed, even though I don't know how heavy the creation process actually is. Likely it just won't matter but this concern also interacts with the mutation-semantics problem above to produce a decidedly non-D3 way of treating these values.

This exposes the fundamental divide between Vue's reactivity and D3's OOPness. I wonder if there's a way to have both: can this mixin be allowed to memoize the values of these objects? I believe that would alleviate both of my general concerns, at the cost of (1) creating extra logic for memoization and (2) creating a risk that this logic will violate Vue's assumptions that underpin the reactivity system.

Let me know what you think of this.

web/src/components/vis/mixins/axisPlot.js
subdavis: axisPlot is not a constructor, so it can't be guaranteed that it will be called before other methods.   For that reason, I'm hesitant to treat it like a constructor.  We don't really have a paradigm for communicating that.

I also don't see why this mixin couldn't be used where the instantiation component managed a couple of plots.  For example, two plots where the domain and range of the data will be identical.

web/src/components/vis/VisPca.vue
subdavis: I don't think this is correct?  After a transformation, `domain` and `range` may change, so we need to re-run the whole `axisPlot` method (which is done on the next line).  that method calls these two at the end, so these aren't necessary.

web/src/components/vis/mixins/axisPlot.js
waxlamp: But can't it be guaranteed that a given user of this mixin can arrange to explicitly call `axisPlot()` before ever calling e.g. `setXLabel()`?

I like the idea of getting the mixin to do double duty, but it gets dicey because `margin` for instance doesn't vary, while the target SVG element would (and presumably the two plots you're managing could in principle have different margin values). My feeling is that if we want to be able to drive two plots within a single component, we'd want to design this mixin to explicitly be able to do that, rather than just pass an SVG element to the utility methods without covering the other aspects of plot management.

web/src/components/vis/mixins/axisPlot.js
subdavis: That argument makes sense to me.  If that's the case, I think this mixin could be made safer by leaning on the lifecycle.

1. Set `axisX, axisY, scaleX, scaleY` as null in mixin data
2. Override whatever data you want from the component 
3. Inside the `mounted()` hook of the mixin (which will be called before the instantiated component) use the data + computed properties to set those 4 mutables, so they're ready by the time you call `axisPlot`.  In `axisPlot`, check that they're ready and throw some error if they aren't.

... interrupted by chat.  Will continue thoughts soon.

subdavis:
BTW;

```js

        loadings: {
          data: null,
          valid: false,
        },
```

neeeded on store/index.js line 99.

web/src/components/vis/mixins/axisPlot.js
waxlamp: This turns out to be harder than I thought. I am going to leave it as-is for now, and if it causes trouble in the future we can revisit.

web/src/components/vis/VisPca.vue
waxlamp: (We discussed this in person: the solution we reached is to rely on D3's imperative functionality for setting axis labels, and remove the reactive properties from the pipeline altogether.)

web/src/components/vis/mixins/axisPlot.js
waxlamp: (also discussed this in person; decided to let `axisPlot` capture the SVG element to avoid having to leak the SVG element in calls to the set label functions.)

web/src/components/vis/mixins/axisPlot.js
waxlamp: I made things `null` that are required to be sent from the base component, and put in reasonable defaults for the rest.

waxlamp:
My changes/additions are ready for review. Thanks again @subdavis.

subdavis:
LGTM @waxlamp !

PR #131, opened 2019-06-24
Change Upload Behavior: error handling + immediate upload

subdavis:
Screenshot_20190624_140838

### DIFF
--- a/web/src/components/Pretreatment.vue
+++ b/web/src/components/Pretreatment.vue
@@ -44,7 +44,7 @@ export default {
 </script>
 
 <template lang="pug">
-v-layout.pretreatment(row, fill-height)
+v-layout.pretreatment-component(row, fill-height)
 
   v-navigation-drawer.navigation(floating, permanent, style="min-width: 220px; width: 220px;")
     v-layout(column, fill-height)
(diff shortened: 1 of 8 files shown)


subdavis:
Working on some more changes here FYI.

subdavis:
## Delete confirmation

Screenshot_20190625_094209

## New upload screen.

Returning to this screen after upload will show all the currently available datasets.

Screenshot_20190625_094148

subdavis:
> After I upload a file, I see a non-zero size listed while the file is being processed; upon completion, the size changes to read 0 B.

File size needs to be added to the rest API.  I can know its size at upload time, but when I query the server for a given dataset, that info is missing.  Should make this into a new issue.

> If a file has semantic issues (e.g., no primary key column), that fact is presented upon moving to the "view data" screen; would it be a good idea to present the problem in icon+popover form alongside the "processed successfully" badge?

I thought about that too.  I got into the weeds a bit, about showing a distillation of the parsing issues.  I suppose this could be boolean.  Either the file is perfect, or it needs cleanup in some way.  Anything more descriptive than this would probably make for clumsy UI.

I'll add this!

> Is the continue button necessary, now that the "view data" buttons are (1) easier to get to with the mouse and (2) take the user directly to the dataset they wish to see?

I believe it is, but could be wrong.  For new / unfamiliar users, I think a large part of this application's value is the guidance it provides.  I think we should raise this question at the meeting today.

subdavis:
@waxlamp 

Screenshot_20190625_132350

waxlamp:
Looks great!

PR #135, opened 2019-06-25
Return raw file size with csv_file response

### LINKED ISSUE #133: Return dataset csv raw size from server
In bytes.

jbeezley:
@subdavis Is this what you are looking for?

Closes #133


subdavis:
Yeah, it is ty

PR #139, opened 2019-06-27
sessionStore and router fixes

subdavis:
## New feature: localStorage sessions

When you upload a CSV, a reference to it will be stored in localStorage.  When you return to this application, those references will be read and their data loaded from the server.  

This PR adds a new "ready" state to each dataset, allowing for a new state where localStorage has been read and loaded into vuex, but the actual data for the dataset has yet to return from server.  

## Bug fixes

All the bugs associated with routing and the expansion list should be resolved here.  This includes the issue with moving between datasets whose transform views are open.

## Notes

This moves the `dataset.source` property to `dataset._source`.  It is a raw copy of the response from the server without any validation or processing, so using it should be avoided in most cases (I think the plots are the exception, but for all other uses, I've moved the expected properties from within source to direct properties of the dataset, like `id` and `name`.

### DIFF
--- a/web/src/store/index.js
+++ b/web/src/store/index.js
@@ -166,7 +166,6 @@ const mutations = {
   }) {
     const { last, ranges, type } = state.datasets[key].selected;
     state.datasets[key].selected.last = idx;
-    // this.selected.last = idx;fplo
     if (event.shiftKey && axis === type) {
       ranges.add(last, idx);
     } else if (event.ctrlKey && axis === type) {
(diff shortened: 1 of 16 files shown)


web/src/store/index.js
```diff
@@ -136,7 +166,7 @@ const mutations = {
   }) {
     const { last, ranges, type } = state.datasets[key].selected;
     state.datasets[key].selected.last = idx;
-    // this.selected.last = idx;
+    // this.selected.last = idx;fplo
```
waxlamp: Can we just strike this line?

subdavis: Yep, done.

PR #140, opened 2019-06-27
[WIP] Denormalize row names and column headers

jbeezley:
This reduces the number of queries needed for most operations internally.  The data consistency is handled by ORM events.  I think this helps performance with less reliance on caching, but I still need to do some actual comparisons between the branches.

The test suite has not yet been updated.


jbeezley:
After some exploration, I don't think this will save any real time.  At best, it will save on the need for some of the caching, but I don't think it will be worthwhile to do.

PR #147, opened 2019-07-07
Remove per-column stats from responses

jbeezley:
The web client currently does not use this information.  It saves 37% of the time required to serialize large tables.


jbeezley:
I removed another property from the serialized output to save another 20%, now over 2x faster than before.  The bulk of the serialization time is now taking the table validation code.

PR #149, opened 2019-07-09
Fix divide-by-zero errors in correlation function

### LINKED ISSUE #142: Divide by zero error
A divide by zero error occurs in some datasets.  We should handle this case explicitly.  See <https://sentry.io/organizations/kitware-metabolomics/issues/1087122843>.

jbeezley:
Fixes #142

### DIFF
--- a/metabulo/views.py
+++ b/metabulo/views.py
@@ -361,13 +361,19 @@ def _get_loadings_data(csv_file):
         return x * x
 
     def cor(xs, ys):
+        if xs == ys:
+            return 1.0
+
         x_mean = mean(xs)
         y_mean = mean(ys)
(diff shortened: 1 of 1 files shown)


jbeezley:
Yeah, it really should be a < epsilon, but I couldn't decide what the appropriate epsilon should be.  e.g. if the mean is 1e-50, a 1e-55 variance can be (but is not always) numerically meaningful.  The case when the arrays are exactly the same is the important thing to catch here.

waxlamp:
Computational numerology is hard :stuck_out_tongue: 

If this fixes the current problem, then it feels like we should go with it for now at least.

PR #151, opened 2019-07-09
Improve the loadings plot

waxlamp:
The main changes here are:
- render the loadings plot without the vector tails
- render the vectors with only a stroke, in a larger size
- fix tooltip format and layout

### DIFF
--- a/web/src/components/vis/VisLoadings.vue
+++ b/web/src/components/vis/VisLoadings.vue
@@ -121,7 +121,7 @@ export default {
             tooltip.transition()
               .duration(200)
               .style('opacity', 0.9);
-            tooltip.html(`col: ${d.col}<br>x corr: ${coordFormat(d.x)}<br>y corr: ${coordFormat(d.y)}`)
+            tooltip.html(`<b>${d.col}</b><br>(${coordFormat(d.x)}, ${coordFormat(d.y)})`)
               .style('left', `${event.clientX}px`)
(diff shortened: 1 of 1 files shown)


web/src/components/vis/VisLoadings.vue
```diff
@@ -98,52 +96,38 @@ export default {
...
           .attr('cy', this.scaleY(0))
           .attr('r', 0)
+          .style('fill-opacity', 0.001)
+          .style('stroke', 'black')
           .on('mouseover', function mouseover(d) {
             select(this)
               .transition()
-              .duration(200)
-              .attr('r', 4);
+              .duration(500)
```
subdavis: Just curious, what prompted this change?  I checked it out and the half-second animation felt a bit long.  I changed it back to 200 and that feels much more crisp to me.  This may be a case for Judge Tomeck to weigh in on.

waxlamp: Well the substantive thing was to reduce the mouseout animation from 500 to 200 ms; the longer transition felt way too "sticky". I had a vague idea that a longer mouseenter animation served us better somehow but:
1. I'm fine with changing both durations to 200ms, and
2. I agree with asking Judge @jtomeck to weigh in here.

waxlamp: Actually let's just go with 200ms. It does feel better, and any arguments I could make about a slower half of the animation are too weak to really belabor.

PR #152, opened 2019-07-09
Improve the scores plot

waxlamp:
- Draw the points larger
- Display data in tooltip more pleasantly


subdavis:
* Should we move the ellipses behind the data points?  I've noticed it sits on top of them.  That's a super minor nitpick.
* I like the animation speed of these data points.  Whatever we decide, it should probably be the same for scores and loadings.

waxlamp:
When I started drawing the dots with a stroke, it exposed this problem of how the ellipses are drawn. Good catch. I moved the ellipses behind the dots.

PTA(nother)L

PR #154, opened 2019-07-15
Fix/148

### LINKED ISSUE #148: Web client performance issues
When loading large tables, the web client can lag considerably.  For example, [this csv file](https://gist.github.com/jbeezley/3c8c4afa075e82d3608d8c3394eb6806) contains over 1000 columns.  On my desktop, just selecting a column (no server side requests) takes between 1-2 seconds.  On a slow laptop, I can get the UI to freeze up for more than 10 seconds by clicking around.

We should look into only rendering the rows/columns in view.

subdavis:
This is a  proof of concept fix for #148.  There are a large number of sacrifices involved, so I need help deciding what to prioritize.

Screenshot_20190715_113059

This is the highest-performance version I could get, and it's perfectly responsive (20-30ms latency between clicks and style changes on my machine).

Some info:

* Swapped cell-level styles on columns for colgroups, dropping the computation cost from MxN to M+N for updates.  Insert is obviously still MxN.  This has a large flexibility cost, because colgroups can only set style [in limited ways](https://www.w3.org/TR/CSS21/tables.html#columns)
* I've dropped borders around the selected region, and around different types of columns.  Changing cell borders forces a table width recomputation and the whole page layout has to recompute at great cost.
* It will be much more challenging now to make the primary key clickable.  I can still do it efficiently, but I consider that lower priority priority.

The nastiest parts here are
* how the table borders look inside highlighted columns.  For example, the grey borders inside the pink metadata columns.  There's just nothing I'm aware of we can do about this: we would have to trade about a second of page unresponsiveness to make that better in the worst case.  
* That I can't set many kinds of column-level styles anymore.  Notice the rows text are white but the columns are black.  colgroups don't let you set cell styles.
* That now, selected columns and rows that already have some formatting are super hard to distinguish.  Example below: the header row is highlighted, but you can't tell.

Screenshot_20190715_114038

cc @jtomeck @jeffbaumes @jbeezley all your thoughts and suggestions are welcome.

fixes #148

### DIFF
--- a/web/src/components/Cleanup.vue
+++ b/web/src/components/Cleanup.vue
@@ -13,41 +13,80 @@ import {
 import { base26Converter } from '@/utils';
 import SaveStatus from '@/components/SaveStatus.vue';
 
+function updateTable(el, binding) {
+  const { dataset, id, activeClasses, setSelection } = binding.value;
+  const colgroup = el.getElementsByTagName('colgroup')[0];
+  const body = el.getElementsByTagName('tbody')[0];
+  const columns = colgroup.children;
+  const rows = body.children;
+  for(let index = 0; index < (columns.length - 1); index += 1) {
+    const col = columns[index + 1]; // Account for 0th being empty
+    col.classList.remove('first', 'last', 'active', 'key', 'group', 'metadata', 'masked', 'measurement');
+    col.classList.add(...activeClasses(index, 'column'));
+    col.classList.add(dataset.column.labels[index]);
+  }
+
+  for(let index = 0; index < (rows.length); index += 1) {
+    const row = rows[index];
+    row.classList.remove('first', 'last', 'active', 'header', 'metadata', 'masked', 'sample');
+    row.classList.add(...activeClasses(index, 'row'));
+    row.classList.add(dataset.row.labels[index]);
+  }
+}
+
 function renderTable(el, binding) {
   while(el.firstChild) {
     el.removeChild(el.firstChild);
   }
-  const { dataset, id, activeClasses } = binding.value;
+
+  const { dataset, id, activeClasses, setSelection } = binding.value;
   const thead = document.createElement('thead');
+  const colgroup = document.createElement('colgroup');
   const tr0 = document.createElement('tr');
   const th0 = document.createElement('th');
+  const col0 = document.createElement('col');
   tr0.appendChild(th0);
+  colgroup.appendChild(col0);
+
   dataset.column.labels.forEach((col, index) => {
+    const coln = document.createElement('col');
     const thn = document.createElement('th');
     const span = document.createElement('span');
     span.innerText = base26Converter(index + 1);
     thn.onclick = (event) => {
-      binding.value.setSelection({ key: id, event, axis: 'column', idx: index });
+      setSelection({ key: id, event, axis: 'column', idx: index });
     }
-    thn.classList.add(...['control', 'px-2'].concat(activeClasses(index, 'column')));
+    thn.classList.add('control', 'px-2')
+    coln.classList.add(...activeClasses(index, 'column'));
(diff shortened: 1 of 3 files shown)


jeffbaumes:
If it really is tons faster to render (sounds like it is) I'd say to go with it. Could you potentially do a second mostly-transparent table overlay to do selection highlighting?

subdavis:
@jeffbaumes Hadn't considered an overlay, and I'm not super familiar with them.

Is this the sort of process you're proposing?

On selection change, inspect the DOM to get the left and right positions of all the rows/columns involved.   Overlay a div with the same width and height with transparency and borders and the like.

That sounds pretty clever and I like the idea.  I'll give it a try soon.

jeffbaumes:
I was thinking a full duplicate of the table HTML, but styled to be
transparent except for semitransparent cell background to show the
selection. Your idea may be even better though.

jtomeck:
@subdavis I know I mentioned pseudo elements before, but do you think this approach would work? I'm putting it here because I'm not super comfortable with how I'd implement it in cleanup.vue currently.

1. On the table cells:
```
position: relative;
z-index: 1;
```
2. On table cells that are, for example, metadata:
```
[metadata cell selector]:after {
    background: var(--v-accent2-lighten3);
    bottom: -1px;
    content: '';
    display: block;
    left: -1px;
    opacity: .75;
    pointer-events: none;
    position: absolute;
    right: -1px;
    top: -1px;
    z-index: -1;
}
```
This adds an overlay to each cell which is z-indexed behind the text. Let me know what you think. I'm not sure if pseudo elements render faster/slower than the colgroup does.

subdavis:
I'll have to give this a try @jtomeck.  Because of the absolute positioning, I have a feeling the layout computation wouldn't be an issue, so this may be quite efficient.  It will be easy enough to test.  

Thanks for the suggestion!

subdavis:
@jtomeck Here's some more CSS debauchery:

Colgroups can set background safely, and backgrounds dont affect width.

```css
background: linear-gradient(90deg, rgb(23, 147, 248) 0px, rgba(161, 213, 255, 0.3) 2px, rgba(161, 213, 255, 0.3) calc(100% - 2px), rgb(23, 147, 248) 100%);
```

Screenshot_20190716_162325

It looks almost the same as box shadow (which I have now decided I prefer) but works with colgroups.

Obvious issue here is that I row styles always layer on top.  Probably not useful after all, but kind of a neat trick.

subdavis:
@jeffbaumes this is ready to go in.

However, it looks like tox is failing on master, so we can't merge until that gets fixed. 

```
ImportError while loading conftest '/home/brandon/github.com/metabulo/tests/conftest.py'.                                                                               
tests/conftest.py:7: in <module>
    from metabulo import models
metabulo/models.py:20: in <module>
    from metabulo.cache import clear_cache, csv_file_cache, region
metabulo/cache.py:71: in <module>
    region.configure('flask_request_local')
.tox/py36/lib/python3.6/site-packages/dogpile/cache/region.py:416: in configure
    backend_cls = _backend_loader.load(backend)
.tox/py36/lib/python3.6/site-packages/dogpile/util/langhelpers.py:40: in load
    return impl.load()
.tox/py36/lib/python3.6/site-packages/pkg_resources/__init__.py:2433: in load
    self.require(*args, **kwargs)
.tox/py36/lib/python3.6/site-packages/pkg_resources/__init__.py:2456: in require
    items = working_set.resolve(reqs, env, installer, extras=self.extras)
.tox/py36/lib/python3.6/site-packages/pkg_resources/__init__.py:786: in resolve
    raise DistributionNotFound(req, requirers)
E   pkg_resources.DistributionNotFound: The 'webargs' distribution was not found and is required by the application
ERROR: InvocationError: '/home/brandon/github.com/metabulo/.tox/py36/bin/pytest tests'
```

PR #156, opened 2019-07-22
NxM manual update

### LINKED ISSUE #148: Web client performance issues

subdavis:
This PR rivals #154 to fix #148 but is much slower.  The benefit is it looks identical and drops the compute time for the same large CSV from 1800ms second updates to  800ms updates.

I've also encountered a [fascinating bug in Vue](https://github.com/vuejs/vue/issues/8728) that has yet to be patched.  I managed to get a 30% speedup by changing my access pattern:

```js
// before
this.nested.array[index]

// after
const arr = this.nested.array;
array[index]; 
```

Note that in this iteration, I am still reading from reactive properties. 

@matthewma7 suggested in #148 that part of cost may come from turning my response data into observables.  Indeed, this process is costly but still takes <1s and only happens once. After seeing the performance issues with `reactiveGetter`, I tried extracting bits from vuex to prevent reactivity, but I only got another 25% boost -- still quite slow to a user.

At this point, I believe I've explored all the sane options for keeping the cleanup table similar to what @jtomeck has designed.

### DIFF
--- a/web/src/components/Cleanup.vue
+++ b/web/src/components/Cleanup.vue
@@ -12,10 +12,12 @@ import {
 } from '@/utils/constants';
 import { base26Converter } from '@/utils';
 import SaveStatus from '@/components/SaveStatus.vue';
+import CleanupTable from '@/components/Table.vue';
 
 export default {
   components: {
     SaveStatus,
+    CleanupTable,
   },
   props: {
     id: {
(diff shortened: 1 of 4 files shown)


subdavis:
closing in favor of #154

PR #158, opened 2019-07-23
Pin pandas to last release and update pip lockfile

jbeezley:
The CI test failure is due to a new release of pandas requiring lzma.  The standard python containers on CircleCI are missing lzma.  For now, I'm mitigating the issue by pinning pandas to v0.24.

@subdavis This doesn't look like the same error you are having, but *maybe* it will fix it.

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -16,12 +16,18 @@
         ]
     },
     "default": {
+        "alembic": {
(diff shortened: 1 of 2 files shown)


subdavis:
I'll rebase and give it a try, thanks @jbeezley

PR #161, opened 2019-08-13
Normalization modes

jeffbaumes:
Adding these normalization modes:
* Sum: Ensure rows sum to 1000.
* Reference Sample: Ensure rows sum to whatever the reference sample sums to. **This should have as input the sample ID to use as the reference sample, but currently uses the first sample row.**
* Weight/Volume: Given a weight/volume column, scale each row so that each weight/volume is equal to 100. **This should have as input the column name to use as the weight/volume for normalization, but currently uses the first metabolite column.**

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,6 +10,7 @@ metabulo = {editable = true,path = "."}
 
 [requires]
 python_version = "3.7"
+pyparsing = "=2.4.2"
 
 [pipenv]
 allow_prereleases = true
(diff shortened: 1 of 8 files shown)


jeffbaumes:
@subdavis and I talked a bit about this. It appears that only normalization will require inputs (not scaling or transformation), and that single input will for the foreseeable future be a single string. Right now I'd recommend just adding a single normalization_argument string field to the CSVFile model to support this, and passing this through the REST call to the normalize() function.

subdavis:
Taking a look at this now.

PR #162, opened 2019-08-14
New vis layout and tweaks

jeffbaumes:
Trying out a modular VisTile component in which to place visualizations. Below is how it looks on a big monitor.

image

Currently the content area (white part) of a tile is set to 600x600, so visualizations would fill this space. We could decide a different size is a better default, and could also add a "full screen" button to maximize plots. The idea is to have the default size be able to fit a 2x2 grid of plots on the screen on reasonable-sized monitors and laptop screens. In particular for the Transform Table page we want to do some interactive linking between plots so having at least two plots visible at a time is critical.

I also updated all fonts to match the app (including unsetting a d3-axis hard-coded font-family), tweaked the sizes of dots to be consistent across plots, thinned the stroke width, updated the tooltip background to be neutral instead of a new color.

I have a request in to @jtomeck to give us some categorical colors that will go with the application theme to replace the orange/blue.

@waxlamp I haven't thought much about where to put extra plot controls, but there could be another slot for drop-down menu items or a dialog or drawer of controls that appears on click (e.g. a three-dots menu in the vis title bar).


jeffbaumes:
Moved plots to upper left.

image

For options, maybe a temporary tray that expands to the right of the vis?

jeffbaumes:
Trying out empty circles for both plots. I like the lack of occlusion for both plots now. Also decided to go back to 1px stroke width because the colors were too washed out.

image

waxlamp:
Looks good.

subdavis:
Really like the look of this.   Eventually we may be able to extend VisTile with some domain/range controls and the ability to export a screenshot like Metaboanalyst does.

Issue #157, opened 2019-07-22
Batch cleanup operations


subdavis:
I suppose this would involve some kind of dirty state and a button the user would need to click to flush the changes to server.

@waxlamp do I understand correctly?

waxlamp:
That's how I'm picturing it, but the idea didn't originate with me (maybe it was @jeffbaumes). I also believe it would be worthwhile to have a deep discussion about whether we'd want this type of interaction, especially if it means that work could inadvertently be lost (and in turn, whether we'd want to support mechanisms to prevent that from happening).

subdavis:
Discussed during 7/23 meeting:

Decided to break label change operations and table validation apart.  Change requests will be quick so the UI isn't spinning, and validation will happen automatically in the background.

jeffbaumes:
@jbeezley is this resolved now because of #163?

jbeezley:
Yes

PR #163, opened 2019-08-19
Add fast column/row label changes

jbeezley:
This addressed #157.  The client needs to be updated to handle the api change on `PUT /csv/:id/batch/label`.

### DIFF
--- a/metabulo/models.py
+++ b/metabulo/models.py
@@ -326,6 +326,9 @@ class CSVFileSchema(BaseSchema):
             if data.get(key) is not None:
                 return data[key].to_csv()
 
+        if 'table' not in data:
+            return data
+
         data['table'] = data['table'].to_csv(header=False, index=False)
         data['measurement_table'] = get_csv('measurement_table')
         data['measurement_metadata'] = get_csv('measurement_metadata')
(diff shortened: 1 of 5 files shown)


subdavis:
@jbeezley Would it be possible to return the new state of all rows and columns without any of the additional processing?  

[I referenced the issue here](https://github.com/girder/metabulo/issues/114#issuecomment-482208306), and I think the same conceptual problem still exists.

> There's a lot of server logic to be duplicated for operations like primary key change, such as what to change the row/column to that used to be the primary key. I'd rather not update the table view manually because any inconsistency between client and server could cause weird state sync problems.
> Given that, I'd propose a third option where PUT csv/id/lable/batch returns the new { rows: [], columns: [] } and none of the other fields so that the request doesn't do validation but label state is confined to the server. The table view can update quickly and run validation asynchronously.

jbeezley:
Yeah, I just forgot that you had requested that before.

subdavis:
Added changes that might work once the server response is modified.   I haven't tested them, and those commits are really more of a remote `git stash`.  They'll conflict with what I've done over @ #164, but this is less invasive and should merge first.

jbeezley:
@subdavis It appears to work with the latest change.

PR #164, opened 2019-08-21
wip

subdavis:
@jbeezley this is sort of what I was thinking.  Plan is to have the argument string be a primary key.

If the primary key for a row/column changes, I'm not sure what the best thing to do is.   I guess the normalization mode should simply default to the primary key at index 0 if the one provided isn't valid or is null.


subdavis:
@jeffbaumes @jbeezley 

This code works but required me to do some kind odd stuff (like sending a ton more arguments through to the normalize method).  Could someone do a quick pass code review?   I'll fix the tests afterward and we can merge.  Ty.

metabulo/views.py
```diff
@@ -180,10 +180,11 @@ def set_normalization_method(csv_id):
     csv_file = CSVFile.query.get_or_404(csv_id)
     args = request.json
     method = args['method']
-    if method is not None and method not in NORMALIZATION_METHODS:
-        raise ValidationError('Invalid normalization method', data=method)
+    argument = args.get('argument', None)
     try:
+        validate_normalization_method(method, argument)
```
jbeezley: It would be better to do this with webargs.

subdavis: Sure, I can move this call into the validator.

metabulo/normalization.py
```diff
@@ -30,9 +38,9 @@ def sum(table):
...
-def reference_sample(table):
-    return table.iloc[0].sum() * table.div(table.sum(axis=1), axis=0)
+def reference_sample(table, argument):
+    return table.loc[argument].sum() * table.div(table.sum(axis=1), axis=0)
 
 
-def weight_volume(table):
-    return 100 * table.div(table.iloc[:, 0], axis=0)
+def weight_volume(table, argument, sample_metadata):
+    return 100 * table.div(sample_metadata.loc[:, argument], axis=0)
```
jbeezley: This could throw or return NaN's if the metadata is non-numeric or zero with little to no indication to the user what has gone wrong.

Ideally, I think this would be a validation error on `PUT .../normalization`; however, that does lead to a possibility for the normalization argument to be valid before a column/row label change in the cleanup.  I'm inclined to ignore this problem for now because the changes to the data model I'm working on will probably reset the normalization/scaling/transforms completely.

subdavis: That being the case, I'm not super inclined to go down path of creating UI for handling these edge cases if it's going to behave differently in the near future.

Should I close this PR and wait on the model changes?

jbeezley: I think it's okay to merge in and avoid worrying about edge cases.

Speculatively, I'm headed toward keeping the cleanup, imputation, column/row labeling where it is.  Then adding a new model for scaling, normalization, and transforms.  This new model would avoid all of the costly validation and row/column lookups that occur in the cleanup.

Right now, I'm not sure about how to generate the new model... Keeping the API exactly the same would require generating it on GET requests because the validation is done lazily.  I'm probably going to end up with some sort of explicit POST indicating clean up is done.  Whether we move the transformation endpoints to a new resource to make the change explicit on the API is still up in the air.

This is all just to say, the server side of this will just get moved directly to a new model.  The REST API will look mostly the same.  I don't see a reason to make validation lazy post-cleanup stage.  Errors should be thrown in the mutation endpoint.

metabulo/views.py
subdavis: Nevermind, I didn't realize the use_kwargs decorator took all the same kwargs that parser.parse does.  There are no examples explaining that in the docs.

subdavis:
@jbeezley this is ready for review.  Notice I've changed the base to master.  This branch contains #161 so that can be closed.   It was easier to only rebase this.

PR #167, opened 2019-09-04
Add a "validated_metabolite_table"

jbeezley:
This implements the refactor we have discussed.  It essentially checkpoints the old `CSVFile` model post imputation.  The purpose of the new table is to store metabolite tables that are *guaranteed* to be valid, so that rather than perform validation lazily in GET requests, they can be done in the PUT request that performs the mutation.

The API changes here are intentionally minimal.  There really is only one minor change that needs to happen in the client.  That is a call to `POST /api/v1/csv/:id/validate` when the clean up is done (or really any time the client wishes to check if the csv_file state is valid).  On success, this will upsert a row in the new table with the given csv file id.  The functionality of the new endpoint could be placed in `GET /api/v1/csv/:id`, but hiding database mutations in a GET requests feels icky.

A caveat: I made no attempt to invalidate this table automatically if something changes on the source csv_file row.  It would be possible (and not that hard), but I don't think it is necessary currently.

You can try it now by uploading a new csv file and performing the cleanup as before.  Once this is done, manually call the  `POST /api/v1/csv/:id/validate` endpoint.  The visualization page should then work correctly.

I haven't bothered updating the tests or generating a migration, but I will do so once others get a chance to comment on it.

### DIFF
--- a/devops/ansible/site.yml
+++ b/devops/ansible/site.yml
@@ -168,5 +168,6 @@
         FLASK_APP: "metabulo.wsgi"
         SQLALCHEMY_DATABASE_URI: sqlite:///{{ sqlalchemy_database_path }}
         DOTENV_PATH: "/home/gunicorn"
+        OPENCPU_API_ROOT: "{{ opencpu_api_root }}"
       become: true
       become_user: gunicorn
--- a/metabulo/models.py
+++ b/metabulo/models.py
@@ -20,8 +20,8 @@ from werkzeug.utils import secure_filename
 
 from metabulo.cache import clear_cache, csv_file_cache, region
 from metabulo.imputation import IMPUTE_MCAR_METHODS, impute_missing, IMPUTE_MNAR_METHODS
-from metabulo.normalization import normalize
-from metabulo.scaling import scale
+from metabulo.normalization import NORMALIZATION_METHODS, normalize
+from metabulo.scaling import scale, SCALING_METHODS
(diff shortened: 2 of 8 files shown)


waxlamp:
Is anything preventing us from sticking that `POST` call in the client now?

jbeezley:
Not really.  I just wanted to get @subdavis's input on it first.

jbeezley:
I hacked in a call to the new endpoint in a way that I think works.  There are almost certainly more streamlined ways of accomplishing this though.

subdavis:
Looks good to me.  I think unless the intended behavior of the UI changes, this is the way to do it.

What would you consider more streamlined?

jbeezley:
I was thinking something more along the lines of explicitly calling the validation endpoint to check the validation state rather than using the GET request to do it.  It's a minor point because the second call to the validation function is quick because it gets cached.

I'm a little iffy about shoehorning this change into exactly the same API.  We may want to make the client aware of the new model in the future.  For the near term goals of getting to multiple dataset analysis, I think it is okay.

jbeezley:
Also, to anyone else working on this.  You can try to run `flask db upgrade` to migrate your database, but it will probably be easier just to delete your sqlite file and regenerate it with `metabulo-create-tables`.

subdavis:
Ok, I want to spend some more time thinking about this.  I don't think I understand the new state model, so I probably just need to draw it out.

PR #169, opened 2019-09-06
Enable principal component selection in visualizations

waxlamp:
This works, but feels like the code infrastructure could be improved.

There is also now a question of whether the PC selector should be at a higher level so that both the score plot and the loadings plot can be yoked together.

Thanks to @subdavis for helping me conceptualize this feature, and to @jtomeck for help with Vuetifying the menu content.


web/src/store/index.js
```diff
@@ -226,7 +226,9 @@ const actions = {
 
   async [LOAD_PLOT]({ commit }, { dataset_id, name }) {
     try {
-      const { data } = await CSVService.getPlot(dataset_id, name);
+      const { data } = await CSVService.getPlot(dataset_id, name, {
+        max_components: 6,
```
jeffbaumes: This is a magic number. At minimum probably should pull out as a constant with a relevant name.

web/src/components/vis/ScorePlot.vue
```diff
@@ -82,6 +82,19 @@ export default {
      *   labels: Object<string, Array<string>>
      * }}
      */
```
jeffbaumes: Seems like this comment is supposed to go directly above `rawPoints` but has been separated by inserted code.

metabulo/views.py
```diff
@@ -325,9 +325,12 @@ def _apply_transforms(csv_file):
...
 
 
-def _get_pca_data(csv_file):
+def _get_pca_data(csv_file, max_components=2):
     table = _apply_transforms(csv_file)
-    max_components = int(request.args.get('max_components', 2))
+
+    max_components_arg = request.args.get('max_components')
+    if max_components_arg is not None:
+        max_components = int(max_components_arg)
```
jeffbaumes: The one-liner in the existing code:

```
max_components = int(request.args.get('max_components', 2))
```

seems to do everything the changed code does. Since no one is using the function keyword argument, why was it added?

web/src/components/vis/VisTile.vue
```diff
@@ -14,6 +20,14 @@ v-flex(shrink=1)
   .white.rounded
     v-toolbar.primary.darken-3.top-rounded(dark, flat, dense)
       v-toolbar-title {{ title }}
+      v-spacer
+      v-toolbar-items
+        v-menu(v-if="hasMenu" bottom offset-y left :min-width="150" :close-on-content-click="false")
```
subdavis: My only thought is that it might be preferable to not automatically wrap this in a menu.  If the controls are simple and small, can they be inlined into the toolbar?

IMO fewer clicks is better clicks.  Users aren't usually too curious about kabob menus.

web/src/store/index.js
waxlamp: And more generally, it should probably be user-specifiable.

metabulo/views.py
waxlamp: One of the commits in the sequence did add a keyword argument to this function, to allow me to specify `max_components` from an internal invocation of the function (when usually you have to send the value in through the request args).

I will fix this on this branch of course, but I think a better solution here is to use webargs to inject query param arguments into the flask routes themselves, then make sure these internal functions simply have keyword args with valid defaults rather than asking for the request args from them.

web/src/components/vis/VisTile.vue
waxlamp: The idea would be to put the slot directly under the `v-toolbar-items`, then make each tile specify `v-menu` or whatever else it wants there?

subdavis: Exactly.  The upside is it lays out your controls in a more immediately visible way.   The downside is it has the potential to look less organized or even messy, maybe.   I actually envision these controls looking a bit like toolbar controls a la Microsoft Office

Screenshot from 2019-09-09 08-00-43

That way the icons are of a unified width and style, but you could use different icons for different plots as appropriate.  I think you'd have a lot of overlap, probably.  Especially if you wanted to do something like allow zoom and pan of the plot.

Now that I think about it, this would be easy to change later.  I'm fine with leaving it as-is now instead of speculating about what additional controls are needed in the future.

waxlamp: @jtomeck had a similar thought re not using a dropdown here.

I am going to file a new issue to look at a different/better layout for the menu controls and leave it as a dropdown for now.

web/src/components/vis/ScorePlot.vue
waxlamp: See 914fe8a2d4496fa9e48dd6d90d7f1789f52442ec.

metabulo/views.py
waxlamp: See 4f7a359c6b13b8263a8d4a510268dac404153ce6 and #170.

web/src/components/vis/VisTile.vue
waxlamp: See 7a47bd71a27770fc3273ae8b09d28d5125a95691 and #172.

web/src/store/index.js
waxlamp: See ab7e4762bae7f5e0df9ec5178d74180392aface5..a3383c3cb98f8334db179e24b6bae12fa3f5701b. These commits abstract out the magic number, and then bind that value to a UI control.

waxlamp:
Thanks for the feedback everyone. Ready for another look.

waxlamp:
Ready now!

waxlamp:
> You went a bit farther with the number of components than I was anticipating. Instead of adding UI for setting the max number of components, could we just let the component number go all the way to the highest possible (which would be something like the number of rows/cols in the table)? Having a UI component for setting the max value for other spin boxes does not seem to add a lot of value.

I agree that not having this control would be best; I was just worried about the performance of doing a "full" PCA. But I tried it just now and for data_original.csv, the time for 6 components is ~200ms, and for 63 components (the actual maximum), it's only ~600ms. So for now I will dump the max_components control and just fire off a request for the max number of components.

waxlamp:
@subdavis and I discussed the UI control placement, and though it's moot if we dump the max components control, I suggested that the PCA visualizations should go under their own "meta tile" grouping them together, and any global controls would go in that tile's toolbar.

It may still make sense to group these visualizations together but we will leave that to a future PR, perhaps involved @jtomeck.

@subdavis's suggestion of using `max(x, y)` as the `max_components` parameter would be another way to get rid of the control, but since it turns out the performance isn't abysmal for asking the max number of PCs upfront anyway, we can just do that for now.

subdavis:
@waxlamp done.

>  but since it turns out the performance isn't abysmal for asking the max number of PCs upfront anyway, we can just do that for now.

I would want to test on our `large.csv` 1000+ column example before drawing that conclusion.  PCA and loadings take several seconds to compute for max_components=2, so I don't know how that would scale.

jbeezley:
This is the large file:
```
$ time curl https://metabolomics.kitware.com/api/v1/csv/ea28ebc3-843e-485b-ac69-a27df203f4f2/plot/pca?max_components=1 > /dev/null

real	0m0.858s
user	0m0.031s
sys	0m0.004s

$ time curl https://metabolomics.kitware.com/api/v1/csv/ea28ebc3-843e-485b-ac69-a27df203f4f2/plot/pca?max_components=10 > /dev/null

real	0m0.895s
user	0m0.032s
sys	0m0.003s
```

At least when the cache it hot, the timing is essentially the same.  In this case, the SVD is pretty quick because the table has a small number rows.

PR #173, opened 2019-09-10
[WIP] Address some layout bugs, add "scroll to column on click"

subdavis:
Some of the minor fixes that have been accumulating.


subdavis:
No longer relevent

PR #177, opened 2019-09-11
fix scrollbars just around table

sgratzl:
in the current version a large cleanup table scrolls the whole page. This will limit it to the table.


subdavis:
@sgratzl thanks for this!  I'd like to make #180 a part of this change.

If you approve that and merge it here, I'll approve this.

PR #186, opened 2019-09-13
Refactor plot recomputation into mixin

subdavis:
This PR moves the logic for reloading plot data into a mixin that should be attached to some component container for the actual visualization element.

You can access the plot's always-up-to-data data by `plot.data`, and the only requirement is the presence of `id` as a property.

Still working on this, need to fix the validation flow.

### DIFF
--- a/web/src/components/vis/LoadingsPlotTile.vue
+++ b/web/src/components/vis/LoadingsPlotTile.vue
@@ -1,12 +1,15 @@
 <script>
-import LoadingsPlot from '@/components/vis/LoadingsPlot.vue';
-import VisTile from '@/components/vis/VisTile.vue';
+import LoadingsPlot from './LoadingsPlot.vue';
+import VisTile from './VisTile.vue';
+import { plotDataMixin } from '../../utils/mixins';
 
 export default {
   components: {
     LoadingsPlot,
     VisTile,
   },
+
+  mixins: [plotDataMixin('loadings')],
 
   props: {
     width: {
@@ -19,8 +22,8 @@ export default {
       type: Number,
       validator: Number.isInteger,
     },
-    points: {
-      validator: d => d === null || Array.isArray(d),
+    id: {
+      type: String,
       required: true,
     },
   },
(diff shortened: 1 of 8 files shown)


subdavis:
Actually, for the sake of time and separating concerns, let's make these two separate PRs.   This one is ready for review.

@waxlamp would you mind this going in before the scree plot?  I'll gladly rebase your branch.

web/src/components/vis/LoadingsPlotTile.vue
```diff
@@ -19,8 +22,8 @@ export default {
       type: Number,
       validator: Number.isInteger,
     },
-    points: {
-      validator: d => d === null || Array.isArray(d),
+    id: {
```
sgratzl: why not adding the `id` property to the mixin?

web/src/utils/mixins.js
```diff
@@ -24,6 +26,48 @@ const loadDataset = {
   },
 };
 
+/**
+ * Expects `id` to exist on the wrapping component
+ * @param {String} plotName name of the plot to subscribe to
+ */
+const plotDataMixin = plotName => ({
```
sgratzl: why do you prefer

```js
const xx = {...};

export {
 xx
}
```

over 

```js
export const xx = {..}
```

subdavis: I like to be able to look at all the exported symbols from a file in one place.  I would probably have written `export const` if it were a constant instead of a function.  

However, I've never really thought too much about it.   Is there a reason to do it the other way?

web/src/components/vis/LoadingsPlotTile.vue
subdavis: Because `id` doesn't need to be a prop for this mixin to work.  `id` could be a computed value or even just a data property.

web/src/utils/mixins.js
sgratzl: no big reason, just out of curiosity. imo the advantage is that if you add a function, you don't have to edit two places

PR #192, opened 2019-09-16
Make LoadingsPlot render method reactive on input

waxlamp:
This works by making the `update()` method, which computes the D3 logic for drawing the chart, into a computed property, and placing a dummy element in the template that depends on the value coming from this property.

### Questions (for anyone to answer, but mainly directed at @subdavis)
1. Is this a good approach to more seamlessly bridging between D3 and Vue?
2. Is there a better tactic than placing a `display: none` element in the template?
3. Is there now a better name than "update" for the D3 render method?

### DIFF
--- a/web/src/components/vis/LoadingsPlot.vue
+++ b/web/src/components/vis/LoadingsPlot.vue
@@ -12,6 +12,7 @@ div
         line.vert
       g.plot
   .tooltip(ref="tooltip")
+  span(style="display: none") {{ update }}
 </template>
 
 <style scoped lang="scss">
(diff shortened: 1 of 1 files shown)


waxlamp:
> I'm fine with this. It's sort of a step toward something like https://medium.com/@simonwuyts/vue-js-and-d3-a-chart-waiting-to-happen-e3ac92b00868
> 
> Worth noting that while it doesn't trip the sensor for https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/no-side-effects-in-computed-properties.md, it seems to be in the spirit of doing so. I don't think that's an issue right now.
> 

Yeah, this could become problematic if we don't establish custom rules for using this technique. Essentially, since I've arranged for this computed property to return a constant value, for the sole purpose of triggering D3 updates on changes to its dependencies, it's simply not useful to make use of the property *value* anywhere, and hopefully that's enough of a signal to other programmers about what's going on here.

> Something I've been thinking about with regards to this style of plot generation is the heft of the animations. Right now, the animations from one position to another are animated with JS in an animation loop, and d3 kinda hides that from you -- is that correct? If we were to do something more like
> 
> ```
> template
>   svg
>     ...
>     circle(x='', y='', ...)
> ```
> 
> then transitions could be done with CSS for better performance.

If D3's animation performance becomes a real problem then we should revisit this; however, I'll add that D3's transition API is simple and powerful, and hooks in with the general D3 style of binding data to visual encodings, etc. That is, if D3 is a core technology here, then we get quite a bit out of using its full power, and I think we'd need a compelling reason to do a sort of hybrid Vue/D3 thing along the lines in that medium article. I'm not opposed to doing that, but I'd need to be convinced.

PR #194, opened 2019-09-16
Simplify score plot props

waxlamp:
This used to take a single object full of props and other cruft, which has now been split out into simpler, more transparent props with more restricted, simpler types.


web/src/components/Transform.vue
```diff
@@ -119,6 +119,7 @@ v-layout.transform-component(row, fill-height)
         score-plot-tile(
             :width="600",
             :height="600",
+            :columns="dataset._source.columns",
```
subdavis: I'm deprecating `_source` soon, and you can get the same thing through an "officially sanctioned" property anway.  Please use `dataset.column.data`

web/src/components/vis/ScorePlot.vue
```diff
@@ -91,23 +91,29 @@ export default {
...
+    },
+    eigenvalues: {
+      required: true,
+      type: Array,
+      validator: prop => prop.every(Number.isFinite),
+    },
+    columns: {
+      required: true,
+      type: Array,
+      validator: prop => prop.every(column => ['column_header', 'column_type'].every(key => Object.prototype.hasOwnProperty.call(column, key))),
```
subdavis: I've never given it much though before, but I'm cautions of O(N) validators.  I'm pretty sure validators don't run in production since they only produce warnings.

web/src/components/vis/ScorePlotTile.vue
```diff
@@ -44,6 +48,32 @@ export default {
     pcY() {
       return Number.parseInt(this.pcYval, 10);
     },
+
+    pcCoords() {
+      return this.maybeData('x', []);
+    },
```
subdavis: I think this is the correct way to handle potentially null props.

web/src/components/vis/ScorePlotTile.vue
```diff
@@ -44,6 +48,32 @@ export default {
...
+      return this.maybeData('labels', {});
+    },
+
+    eigenvalues() {
+      return this.maybeData('sdev', []);
+    },
+  },
+
+  methods: {
+    maybeData(key, dflt) {
```
subdavis: There's already a method for this on the plotData mixin called `getPlotDataProperty`.  It doesn't do default values but it could be made to.  I think this is generic enough that we can let the mixin do it.

web/src/components/vis/ScorePlotTile.vue
```diff
@@ -44,6 +48,32 @@ export default {
...
+      const {
+        plot,
+      } = this;
+
+      return plot.data ? plot.data[key] : dflt;
```
subdavis: I do this a lot myself, but it's technically not correct.  If your actual value is falsey and your default value is truthy, this will produce the wrong output.  Better to check if `plot.data[key] === undefined`.

I used to be more rigorous about it, then I settled into using this and `value || default`.

web/src/components/vis/ScorePlotTile.vue
```diff
@@ -54,7 +84,11 @@ vis-tile(v-if="plot", title="PCA Score Plot", :loading="plot.loading")
   score-plot(
       :width="width",
       :height="height",
-      :raw-points="plot.data",
+      :pc-coords="pcCoords",
+      :row-labels="rowLabels",
+      :group-labels="groupLabels",
+      :eigenvalues="eigenvalues",
+      :columns="columns",
       :dataset="dataset",
```
subdavis: Looks like dataset was removed from the score-plot props.

web/src/components/vis/ScorePlot.vue
subdavis: This was just a musing, and can be ignored.

waxlamp: Adopting typescript will moot this concern in any case.

web/src/components/vis/ScorePlotTile.vue
waxlamp: I will extend that method to handle default values and replace the invocations to `this.maybeData()` here.

web/src/components/vis/ScorePlotTile.vue
waxlamp: I might be missing something: this checks to see whether `plot.data` is falsy, not `plot.data[key]`. My understanding is that if the plot data isn't ready yet, `plot.data === null` holds, and this method would therefore return `dflt`. If `plot.data` has an actual object in it, then `plot.data[key]` comes back as the result, regardless of its truthiness. No?

subdavis: Yep, you're right.

web/src/components/Transform.vue
waxlamp: See b2a6c5ae5ec83409a3883c6f68bdc55803744f13.

web/src/components/vis/ScorePlotTile.vue
waxlamp: See aab86b7fc83ed78e80e122d4b956a573fc59a190. I'm leaving this conversation unresolved in case I did something wrong here.

web/src/components/vis/ScorePlotTile.vue
waxlamp: (I am still curious about this topic in general, but it's mooted by using the plotData mixin's data grabbing method.)

web/src/components/vis/ScorePlotTile.vue
waxlamp: See 2b7a292eddc114ddcbc25e20faa21938ff0c2bb8.

PR #195, opened 2019-09-16
Make score plot more reactive

waxlamp:
Depends on #194.

This makes the score plot more reactive in the same manner as #192 did for the loadings plot.


waxlamp:
Closing in favor of #210

PR #196, opened 2019-09-16
Excel Upload

### LINKED ISSUE #190: Import from Excel
When an excel file is detected, let the user choose the excel worksheet(s) to import and upload each as a file in the list.

sgratzl:
based on #188

closes #190 

allows the user to upload an Excel file. Each non empty worksheet will be automatically imported. 

e.g. 
[test.xlsx](https://github.com/girder/metabulo/files/3617583/test.xlsx)

notes
 * please run `pipenv install` to install the missing dependency


metabulo/views.py
```diff
@@ -86,6 +89,55 @@ def upload_csv_file(file, meta):
...
+        if len(excel_sheets) == 1:
+            # single file case, simple name
+            data = next(iter(excel_sheets.values()))
+            push_file(basename, data)
+        else:
+            for sheet, data in excel_sheets.items():
+                push_file(basename.with_name('%s-%s' % (basename.name, sheet)), data)
+        db.session.commit()
+
+        return jsonify([_serialize_csv_file(f) for f in db_files]), 201
```
jbeezley: I usually like to serialize before committing so if there is an error the database gets cleaned up.

sgratzl: make sense, I copied the structure from `upload_csv_file`

metabulo/views.py
```diff
@@ -86,6 +89,55 @@ def upload_csv_file(file, meta):
...
+@use_kwargs({
+    'file': fields.Field(location='files', required=True,
+                         validate=lambda f: f.content_type in _excel_mime_types),
+    'meta': JSONDictStr(missing={})
+})
+def upload_excel_file(file: FileStorage, meta: Dict):
+    excel_sheets = pandas.read_excel(file, sheet_name=None)  # type: Dict[str, pandas.DataFrame]
+    excel_sheets = {sheet: data for (sheet, data) in excel_sheets.items() if not data.empty}
+
+    basename = PurePath(cast(str, file.filename)).with_suffix('')
```
jbeezley: What's the purpose of the type casting?  Isn't `file.filename` already a string?

web/src/components/Upload.vue
```diff
@@ -75,12 +75,15 @@ export default {
         progress: {},
         meta: {},
       })));
+
+      const isExcelFile = file => file.name.match(/\.xlsx?$/i);
```
jbeezley: There probably isn't a real difference, but I would use `file.type` to match the mimetype instead since that's what the server is relying on.

metabulo/views.py
sgratzl: it should be but according to mypy and the typings is isn't. Thus the type cast to satify the mypy linter.

jbeezley: Sounds good.  At some point, I need to figure out how to get type validation working in my editor.

metabulo/views.py
sgratzl: done

sgratzl:
> Could you add a small test case for the new endpoint?

done

web/src/components/Upload.vue
sgratzl: I added both since you cannot fully trust the mimetype in different operating systems

https://developer.mozilla.org/en-US/docs/Web/API/File/type
```Developers are advised not to rely on this property as a sole validation scheme.```

metabulo/views.py
sgratzl: note: I had to undo it, cause the serialization didn't work properly (like it couldn't identify any columns anymore)

jbeezley: Does it work if you run `db.session.flush()` before serialization?

jbeezley:
I'll go ahead and merge this and my `validate-upload-mimetype` branch.

PR #210, opened 2019-09-17
Make score and scree plots more reactive

waxlamp:


web/src/components/vis/ScreePlot.vue
```diff
@@ -164,35 +165,7 @@ export default {
...
-  },
 
-  mounted() {
-    const svg = select(this.$refs.svg);
-    this.axisPlot(svg);
-
-    this.setXLabel(this.xlabel);
-    this.setYLabel(this.ylabel);
-
-    this.update();
```
subdavis: I think this line is still necessary.    Right now, the plot sort of depends on the race condition that some network request will take long enough that a vuex mutation will come along and trigger recomputation some time after the mounted hook.

But this component isn't vuex-aware, so if all the necessary props were already set to their final values at the time of mount, there would be no later trigger to cause the visualization to display.

waxlamp: See 9441a2e.

subdavis:
My :+1: stands.  Merge when ready :)

Issue #185, opened 2019-09-13
ANOVA


jeffbaumes:

subdavis:
For [this file](https://drive.google.com/open?id=1u_y19Bz0rELgizxbBLSvlOwlBnd_Dp4G) Anova throws a 400 with the message below.  It happens when you resolve the errors but not the warnings.

```
residual sum of squares is 0 (within rounding error)

In call:
Anova.lm(mod, type = "III")
```

jbeezley:
We should come up with a standard way of handling errors raised during the processing of analyses so we can show something meaningful to the user.

subdavis:
I may be wrong here, but I think any error raised during an analysis is a bug that we need to fix.  If a dataset passes validation, it should be ready for every analysis.  If that isn't feasible, we probably need a way of indicating to the client that a particular analysis isn't valid for a dataset.

That isn't to say we shouldn't handle errors gracefully, but we should probably still throw those to sentry and fix them.

jbeezley:
> I may be wrong here, but I think any error raised during an analysis is a bug that we need to fix. If a dataset passes validation, it should be ready for every analysis.  If that isn't feasible, we probably need a way of indicating to the client that a particular analysis isn't valid for a dataset.

Ideally yes, but it could easily become intractable to handle this all in the main validation step.

I'm not advocating for blindly ignoring exceptions.  I'm imagining something like the ANOVA analysis running a precheck on the dataset--if the check fails it returns a 400 with a message to display to the user.

> That isn't to say we shouldn't handle errors gracefully, but we should probably still throw those to sentry and fix them.

Agreed, anywhere a 500 is returned from the server, we should show a "Sorry, something went wrong" message.

PR #212, opened 2019-09-17
Refactor analysis, Add Wilcoxon, Add ANOVA

subdavis:
I'm going to try and resolve the plot differences here.

#186, #181, #185 should close in favor of this PR.


subdavis:
@sgratzl there's a good bit of cleanup I still need to do, but the organization here is a good bit more unified.

web/src/components/AnalyzeData.vue
```diff
@@ -0,0 +1,35 @@
+<script>
+import Vue from 'vue';
+import { analyses } from '.';
```
sgratzl: imo not a good style to import from its own `index.js`. Since it is usually the place to export data for other packages

web/src/store/index.js
```diff
@@ -39,17 +41,39 @@ const plotDefaults = {
...
+    args,
+    type,
+  };
+});
+
 const appstate = {
   // map of all datasets in the session by csv UUID
   datasets: {},
   plots: {},
+  analyses: {},
```
sgratzl: still needed?

web/src/components/index.js
```diff
@@ -0,0 +1,28 @@
...
+    path: 'wilcoxon', // same as key
+    name: 'Wilcoxon signed-rank test',
+    shortName: 'Wilcoxon Test',
+    description: 'The Wilcoxon signed-rank test tests the null hypothesis that two related paired samples come from the same distribution. In particular, it tests whether the distribution of the differences x - y is symmetric about zero. It is a non-parametric version of the paired T-test.',
+    component: WilcoxonPlotTile,
+    args: {
+      zero_method: wilcoxon_zero_methods[0].value,
+      alternative: wilcoxon_alternatives[0].value,
+    },
+    type: plot_types.ANALYSIS,
```
sgratzl: isn't it always this plot type?

web/src/components/index.js
```diff
@@ -0,0 +1,28 @@
+import WilcoxonPlotTile from './vis/WilcoxonPlotTile.vue';
+import AnovaTableTile from './vis/AnovaTableTile.vue';
+import { wilcoxon_zero_methods, wilcoxon_alternatives, plot_types } from '../utils/constants';
```
sgratzl: I prefer to have multiple `constants.js` files since with a growing application it will get pretty large, even tho certain elements (like `wilcoxon_zero_methods`) is only used one in the application

web/src/store/index.js
```diff
@@ -76,8 +99,8 @@ function _invalidatePlots(state, { key, plotList }) {
     Vue.set(state.plots[key][name], 'valid', false);
   });
 }
+
 function _setLables(state, { key, rows, columns }) {
-  _invalidatePlots(state, { key, plotList: ['pca', 'loadings'] });
```
sgratzl: why is the plot invalidation not longer needed?

web/src/components/index.js
subdavis: No.  there's also plot_types.TRANSFORM.   Distinguishes which API to call.  I actually don't prefer this, I think it would be better to have a single URI for all plot types.

* `/csv/<id>/plot/pca`
* `/csv/<id>/analysis/anova`

These could both go under the `analysis` route IMO.

web/src/store/index.js
subdavis: Nope, I'll remove that.

web/src/components/index.js
subdavis: Sure, that's fine.  Could you take a look the constants file and enumerate how you'd like to break that up?  How many constants files will we end up with, and what will their paths be?

web/src/components/AnalyzeData.vue
subdavis: Couldn't really come up with a valid place to put that info.  Where would you move that object to?  Perhaps in a new more localized constants.js file?

web/src/store/index.js
subdavis: Because right now changing a label always triggers load_dataset which itself always invalidates plots.  Doing it twice was causing the plots to load before and after.  I plan on refactoring this when I make the `validation` changes mentioned in slack today.

web/src/store/index.js
subdavis: Nope, thanks.  I'll remove it.

web/src/components/index.js
sgratzl: I meant more why not specifying the `type: plot_types.ANALYSIS` when you mix it in the `plotDefaults`. There is no need that this is part of this analysis list

web/src/components/AnalyzeData.vue
sgratzl: I wouuld go for `./vis/analyses.js` .. .default export

web/src/components/index.js
sgratzl: I prefer the localization principle. So, if imported in a single file -> but it directly in the file. Otherwise in a `constants.js` in the most common directory level where it will be imported. e.g. the `wilcoxon_zero_methods` would end up in `./vis/constants.js`

web/src/components/index.js
subdavis: Any new item could be added to the list here and may invoke a different endpoint to get the necessary data.  `data, valid, loading` are all safe to define during the mixin loop because they are properties managed by the vuex state.  type is a constant value intrinsic to the object declared in index.js

web/src/components/index.js
subdavis: In that case, it would probably be best to discuss that in another PR.  That refactor doesn't really affect any of the concerns of this PR.  Would that be alright?

waxlamp: Sounds like we should file an issue around refactoring constants.

sgratzl: -> https://github.com/girder/viime/issues/225

metabulo/views.py
```diff
@@ -529,3 +530,39 @@ def get_pca_overview(validated_table):
     png_content = opencpu.generate_image('/metabulo/R/pca_overview_plot',
                                          validated_table.measurements)
     return Response(png_content, mimetype='image/png')
+
+
+@csv_bp.route('/csv/<uuid:csv_id>/analyses/wilcoxon', methods=['GET'])
+@use_kwargs({
+    'zero_method': fields.Str(missing='wilcox',
+                              validate=validate.OneOf(['wilcox', 'pratt', 'zsplit'])),
+    'alternative': fields.Str(missing='two-sided',
```
waxlamp: Is `two-sided` a specific enough name? Or is that a qualifier for different kinds of tests (e.g., "two-sided t-test")?

metabulo/views.py
```diff
@@ -529,3 +530,39 @@ def get_pca_overview(validated_table):
...
+                              validate=validate.OneOf(['two-sided', 'greater', 'less']))
```
waxlamp: Similar question for the terms `greater` and `less`--these don't sound like standard terms qualifying a test.

metabulo/views.py
```diff
@@ -529,3 +530,39 @@ def get_pca_overview(validated_table):
...
+})
+@load_validated_csv_file
+def get_wilcoxon_test(validated_table: ValidatedMetaboliteTable,
+                      zero_method: str, alternative: str):
```
waxlamp: Are `zero_method` and `alternative` meant to be constrained to the same list of values as in the validation above? In that case the `Literal` type would be good to put in here.

metabulo/views.py
```diff
@@ -529,3 +530,39 @@ def get_pca_overview(validated_table):
...
+@csv_bp.route('/csv/<uuid:csv_id>/analyses/anova', methods=['GET'])
+@use_kwargs({
+    'group_column': fields.Str()
+})
+@load_validated_csv_file
+def get_anova_test(validated_table: ValidatedMetaboliteTable, group_column: Optional[str] = None):
+    measurements = validated_table.measurements
+    groups = validated_table.groups
+
+    group = groups.iloc[:, 0] if group_column is None else groups.loc[:, group_column]
```
waxlamp: Is it ok to assume the first column is the group column? Why not force the client to specify the group column in all cases?

web/src/components/AnalyzeData.vue
```diff
@@ -0,0 +1,35 @@
+<script>
+import Vue from 'vue';
+import { analyses } from './vis';
+
+export default Vue.extend({
```
waxlamp: Why this `Vue.extend()` idiom here, instead of just exporting the object?

web/src/components/AnalyzeData.vue
```diff
@@ -0,0 +1,35 @@
...
+  props: {
+    id: {
+      type: String,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      analyses,
```
waxlamp: Is there a reason this is not passed in as a prop?

web/src/components/Pretreatment.vue
```diff
@@ -13,6 +14,7 @@ export default {
   data() {
     return {
       datasets: this.$store.state.datasets,
+      analyses,
```
waxlamp: Related to my last question, it seems like we're repeating some data storage in different places.

web/src/components/ToolbarOption.vue
```diff
@@ -0,0 +1,31 @@
+<script>
```
waxlamp: This should maybe be called `RadioGroup` or `OptionGroup` or something like that.

web/src/components/vis/AnovaTable.vue
```diff
@@ -0,0 +1,64 @@
+<script>
+import { format } from 'd3-format';
+
+export default {
+
+  props: {
+    data: {
+      validator: prop => !prop || ('data' in prop && 'pairs' in prop),
```
waxlamp: My limited personal experience suggests that this should be split into two props that can be type-validated and named in the code more easily than a composition of that same data into an object.

waxlamp: (e.g., the `items` and `pairs` computed properties below could be eliminated altogether.)

web/src/components/vis/AnovaTable.vue
```diff
@@ -0,0 +1,64 @@
...
+    items() {
+      return (this.data && this.data.data) || [];
+    },
+    pairs() {
+      return (this.data && this.data.pairs) || [];
+    },
+    headers() {
+      return [
+        {
+          text: 'Metabolite',
```
waxlamp: This is probably fine, but we may want to consider keeping these components more general purpose, then specializing them to this application. We may also *not* want to consider that, since we *are* developing a metabolomics application here.

web/src/components/vis/VisTile.vue
```diff
@@ -5,6 +5,10 @@ export default {
       type: String,
       default: '',
     },
+    loading: {
+      default: false,
+      validator: () => true,
```
waxlamp: Doesn't seem like we need a validator here.

web/src/components/vis/WilcoxonPlot.vue
```diff
@@ -0,0 +1,93 @@
+<script>
+import { scaleSequential } from 'd3-scale';
+import { interpolateGreys } from 'd3-scale-chromatic';
+import { format } from 'd3-format';
+
+export default {
+  props: {
+    data: {
+      required: true,
+      validator: prop => !prop || ('data' in prop && 'indices' in prop),
```
waxlamp: Same comment as above about destructuring object props.

web/src/components/vis/index.js
```diff
@@ -0,0 +1,28 @@
+import WilcoxonPlotTile from './WilcoxonPlotTile.vue';
+import AnovaTableTile from './AnovaTableTile.vue';
+import { wilcoxon_zero_methods, wilcoxon_alternatives, plot_types } from '../../utils/constants';
+
+export const analyses = [
+  {
+    path: 'wilcoxon', // same as key
```
waxlamp: I don't understand what the comment means.

subdavis:
Forgot #227 would cause havoc here....   I'll rebase again.

web/src/components/vis/index.js
```diff
@@ -0,0 +1,28 @@
+import WilcoxonPlotTile from './WilcoxonPlotTile.vue';
```
waxlamp: Can you explain why this file is the `index.js` for the `vis` directory?

web/src/components/vis/mixins/plotData.js
```diff
@@ -0,0 +1,54 @@
...
+      dataset() { return this.$store.getters.dataset(this.id); },
+      plot() { return this.$store.getters.plot(this.id, plotName); },
+    },
+    watch: {
+      // eslint-disable-next-line func-names
+      'plot.valid': function () { this.reloadPlot(); },
+      id() { this.reloadPlot(); },
+      active(val) {
+        if (val) {
+          this.reloadPlot();
```
waxlamp: Is this meant to solve the problem of things not updating after making changes earlier in the pipeline? Does this solve the ["sticky vis" problem](https://github.com/girder/viime/issues/215)?

web/src/components/vis/mixins/plotData.js
```diff
@@ -0,0 +1,54 @@
...
+    activated() { this.active = true; },
+    deactivated() { this.active = false; },
+    mounted() { this.reloadPlot(); },
+    computed: {
+      ...mapState(['loading']),
+      dataset() { return this.$store.getters.dataset(this.id); },
+      plot() { return this.$store.getters.plot(this.id, plotName); },
+    },
+    watch: {
+      // eslint-disable-next-line func-names
```
waxlamp: This is to avoid having to name the anonymous function in the next line, right? Is there any reason not to give it a low-key name like `_` or `x` or something?

web/src/store/index.js
```diff
@@ -39,17 +41,39 @@ const plotDefaults = {
...
   },
 };
 
+analyses.forEach(({
+  args,
+  path,
+  key,
+  type,
+}) => {
+  plotDefaults[key || path] = {
```
waxlamp: Why is there a distinction between `key` and `path`? It seems like it would be better to be explicit about this in the `analyses` list, rather than put the mysterious comment there (which I think I now understand :smiley_cat:) and this slippery-feeling logic here.

web/src/components/vis/mixins/plotData.js
subdavis: I had no idea you could name inline functions.

web/src/components/vis/mixins/plotData.js
subdavis: Yes, this should(?) solve that problem.

web/src/components/vis/index.js
subdavis: cc @sgratzl

web/src/components/vis/VisTile.vue
subdavis: You have to have either a validator or a type to pass the linter.   This is the only way I know to accept any value without warnings, and everything in javascript has truthiness.

web/src/components/vis/AnovaTable.vue
subdavis: `items` and `pairs` come from `plot.data.data` and `plot.data.pairs` respectively.  plot.data can be null when the plot is invalid, and so will throw an error about accessing properties on null.  I don't think it should be `Transform.vue`'s problem to deal with this because it would push all these computed props into Transform.vue, but I would be willing to reconsider that position.

web/src/store/index.js
subdavis: cc @sgratzl

web/src/components/AnalyzeData.vue
subdavis: This component is called from the router, so there is nothing to fulfill that prop.

web/src/components/Pretreatment.vue
subdavis: I believe this is just making some constant values available to the template.  This doesn't need to be a reactive object and the template wont change its value, so it could just  as well be a functional component we factor out, but IMO that isn't necessary yet.

devops/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM metabulo/opencpu-base
 
 ADD metabulo /metabulo
-RUN R -e 'library(devtools); library(roxygen2); setwd("/metabulo"); document(); setwd("/"); install("metabulo")'
+RUN R -e 'library(devtools); library(roxygen2); setwd("/metabulo"); devtools::update_packages(devtools::dev_package_deps()$package, dependencies=NA, quiet=TRUE, upgrade="never"); document();  setwd("/");install("metabulo", dependencies=NA, quiet=TRUE, upgrade="never")'
```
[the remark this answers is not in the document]
sgratzl: I start adding dependencies to the R package itself, and this change ensures that they are installed upon installing the R package. Moreover, I deactivated to upgrade existing packages.

metabulo/analyses.py
```diff
@@ -0,0 +1,48 @@
...
+from .opencpu import opencpu_request
+
+
+def wilcoxon_test(measurements: pd.DataFrame, zero_method: Optional[str] = None,
+                  alternative: Optional[str] = None) -> Dict[str, Union[List[str], List[Dict]]]:
+
+    if zero_method is None:
+        zero_method = 'wilcox'
+    if alternative is None:
+        alternative = 'two-sided'
```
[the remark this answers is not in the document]
sgratzl: cause I wanted to be None a valid parameter which simplifies when calling the method with the alternative parameter but with None since no value is given, e.g. from the web request

metabulo/views.py
```diff
@@ -529,3 +530,39 @@ def get_pca_overview(validated_table):
     png_content = opencpu.generate_image('/metabulo/R/pca_overview_plot',
                                          validated_table.measurements)
     return Response(png_content, mimetype='image/png')
+
+
+@csv_bp.route('/csv/<uuid:csv_id>/analyses/wilcoxon', methods=['GET'])
+@use_kwargs({
+    'zero_method': fields.Str(missing='wilcox',
```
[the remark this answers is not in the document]
sgratzl: those are directly taken from https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.stats.wilcoxon.html

web/src/components/vis/index.js
sgratzl: see discussion whether this should be put in an `analysis.js` file

subdavis: sorry, I've moved that to `vis/analyses.js` now

metabulo/analyses.py
```diff
@@ -0,0 +1,48 @@
...
+        (score, p) = wilcoxon(x_data, y_data, zero_method=zero_method, alternative=alternative)
+        data.append(dict(x=x_index, y=y_index, score=score, p=p))
+
+    return {
+        'indices': indices,
+        'data': data
+    }
+
+
+def anova_test(measurements: pd.DataFrame, groups: pd.Series) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
subdavis: We have 1 file for plots called `plot.py`.  It only has pca so far, but if it shouldn't have more, does that need to be renamed pca.py?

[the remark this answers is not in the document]
waxlamp: Under my wiggly feeling above, yes, but that needs to be balanced against the future complexity of doing something like this.

If you're fine with keeping these things confined to one file, we can just forget about this at least for now.

metabulo/views.py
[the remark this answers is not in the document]
waxlamp: Weird that they use "wilcox" instead of "wilcoxon" :angry:

web/src/components/AnalyzeData.vue
waxlamp: Can't we pass that prop in using ["object mode"](https://router.vuejs.org/guide/essentials/passing-props.html#object-mode)?

web/src/components/vis/AnovaTable.vue
waxlamp: If there's a reason to have it this way, I'm ok with it. To do it my way would require yet another `valid` prop to be computed by Transform (as well as conditionally sending in `null` for the putative `data` and `pairs` props), and maybe that's a bridge too far.

I just like the ability to simplify the local logic by splitting structured props apart, but here it looks like it would just shift the complexity to Transform.

web/src/components/vis/VisTile.vue
waxlamp: Maybe a comment explaining why we have an all-pass validator here would be good.

web/src/components/vis/WilcoxonPlot.vue
waxlamp: (Resolved as per similar conversation above.)

metabulo/views.py
subdavis: Offloaded to https://github.com/girder/viime/issues/228

metabulo/views.py
subdavis: Offloaded to https://github.com/girder/viime/issues/228

web/src/components/AnalyzeData.vue
subdavis: fixed

web/src/components/AnalyzeData.vue
subdavis: No, we still need id from router mode.

web/src/components/vis/index.js
subdavis: removed.

metabulo/views.py
subdavis: Offloaded to #228

web/src/components/vis/VisTile.vue
waxlamp: Wait, why not just say this is a boolean value? Surely that's better for code style throughout the uses of this `loading` prop than allowing any truthy/falsy value?

web/src/store/index.js
sgratzl: in the end the `path` has to match the `plotName` (formally known as `key`) used when defining the mixin. We could separate the two or leave it as convention

metabulo/views.py
sgratzl: > first column is the group column
the first column of the list of group columns. 

cause in most cases there is only one group column per dataset thus specifying that by the user/client is redundant

web/src/components/ToolbarOption.vue
sgratzl: `RadioGroup` is too general imo. This one is designed to work precisely in the left toolbar of an analysis.

subdavis: I'll rename this.

web/src/store/index.js
subdavis: Because `key` is no longer a property, I'm just going to change this line to use `path`.

Issue #198, opened 2019-09-17
Data source meta data


sgratzl:
at the current front page there are two drop downs for adding meta data to the data set. However, they are disabled and cannot be edited.

![image](https://user-images.githubusercontent.com/4129778/65041539-0278cf80-d925-11e9-870f-4e6b4472378e.png)

jeffbaumes:
We will eventually want some basic metadata information to help people keep track of their datasets that will get saved back to the DB, e.g. measurement type (NMR, MS, etc.), sample type (serum, urine, etc.). We don't know the exact set of these and allowed values, so we just disabled them for now. I'd be ok with removing them for the time being, I doubt there is much code behind them.

PR #217, opened 2019-09-17
breadcrumb navigation

### LINKED ISSUE #205: UI naming inconsistencies
e.g. `Clean Up Table` in the left side panel vs `Cleanup Data` in the top breadcrumb. 

same for `Transform Table` vs `Transform Data`

### LINKED ISSUE #206: Breadcrumb navigation
right now the top appbar looks like this:

image

an common alternative is to have the following layout: (from left to right)
 * logo + App Title ... clickable to go back to the main screen
 * Breadcrumb ... clickable to navigate and go up in the hierarchy
 * spacer
 * Options (like Profile, General Preferences, ...) on the right side of the bar

sgratzl:
closes #206 
image

and closes #205


subdavis:
> without a meaningful page for the whole data set (like the #198) it will just show the two levels as before

Why couldn't it show more than just 2 routes?  

For example, 
`/pretreatment/cf0a0e50-481f-48d9-9e8e-d774d513f6de/analyze/wilcoxon`

Everything after `/pretreatment/<id>` is a valid route, so `/analyze` and `analyze/wilcoxon` should each have their own route link.   This is appropriate since the breadcrumb will only show on sub-routes underneath pretreatment.

It's worth mentioning that the pretreatment route is, IMO, only a temporary home for analyses.  It was never really intended to be the place where you do everything.  At some point merge operations and analyses should probably move to some other place within the app, or else we should restructure the router to stop calling that route pretreatment.

sgratzl:
I implemented a proper nesting based on nested vue-router routes

web/src/App.vue
```diff
@@ -1,21 +1,42 @@
...
-  v-toolbar.main-toolbar.darken-4(dark, dense, color="primary")
-    v-toolbar-items
-      v-btn.px-0.square-button.darken-4(flat, color="white", to="/select")
-        v-icon {{ $vuetify.icons.close }}
-    v-spacer
-    v-toolbar-title.title
-      span.font-weight-bold.secondary--text.text--lighten-3 Biomarker Analysis »
-      span.font-weight-bold.white--text  {{ $router.currentRoute.name }}
+  v-toolbar.darken-3(dense, dark, color="primary")
+    v-toolbar-side-icon.logo(@click="$router.push('/')") BA
```
waxlamp: This should probably read `VIIME` now that we've settled on the name.

Thoughts on capitalization, @jeffbaumes?

subdavis:
I actually have a general question.  What UX issue does this PR solve?  The information contained in the breadcrumb is also contained in the hierarchy of the tree in the left nav bar.

The workflow was originally intended to have this pretreatment page be an overlay-type page and the analysis take place in some other route.  That's the reason for the X in the upper left: We could pop this page up from anywhere in the UI if you wanted to go back and look at the raw table and tweak the normalizations etc. even if you were in the middle of an analysis.

Maybe @jtomeck would have some thoughts here.  In general, I think this change is small but makes a large impact on the nature of the pretreatment route.   We probably need to think more about how the user moves into the merging and analysis parts of the app.

jeffbaumes:
@subdavis To the point about long-term UI design: I realize this pulls us further away from the original wizard steps design, but this seems to a natural path for the UI to follow, meaning having the whole workflow on the left instead of a sequence of wizard-like steps to switch between. It feels much more intuitive and obvious to me to swap between workflow steps using a sidebar than a forward/back type controls. This also has implications for the pretreatment endpoint naming as you mention elsewhere.

One concern to raise is that this would put us much closer to the left panel UI in metaboanalyst which we do want to differentiate from. But I don't want to differentiate at the expense of good UX, and I don't think anyone would say our UI is a ripoff of that tool since it still looks very different.

As for the repetition (which is a good point): If navigation is specified on the left panel, then do we need a toolbar title at all? Should it just be "VIIME" on the left and options/profile on the right? If eventually we have so much in the left panel that the selection could scroll out of view, the repetition in the toolbar could still be useful/needed.

subdavis:
>  I realize this pulls us further away from the original wizard steps design, but this seems to a natural path for the UI to follow,

As long as we're conscious of the shift and it's intentional, I'm fine with this change.

> But I don't want to differentiate at the expense of good UX 

totally agree.

> if eventually we have so much in the left panel that the selection could scroll out of view, the repetition in the toolbar could still be useful/needed.

overflow is a fair point.  As long as the toolbar is there, the breadcrumb isn't wasting any space: it's just filling up blank space.

---------------

If we're now operating under the assumption that the whole application will be driven from the left navbar, perhaps we should rename the `pretreatment` route  (and the associated component) to `dataset`.

@sgratzl mentioned that there was no overview page for a dataset.  I think it would be nice to have one.

web/src/App.vue
jeffbaumes: Let's do VIIME for now. I've chatted with @jtomeck a bit and it seems there are several routes we could go with capitalization/logo/etc. and I don't think that's the priority right now. We'll do a branding pass later on.

jtomeck:
+1 to the overview page. Let me know if you need some sort of design added to the mockups for that or the sidebar, or if you think you have a clear enough path you can build it and I can provide feedback once it's ready.

sgratzl:
overview page extracted to https://github.com/girder/viime/issues/235

sgratzl:
I changed the title. Anything else we should change in this PR? imo other bigger changes should be extracted to a separate issue.

image

web/src/App.vue
```diff
@@ -1,21 +1,49 @@
...
-      span.font-weight-bold.secondary--text.text--lighten-3 Biomarker Analysis »
-      span.font-weight-bold.white--text  {{ $router.currentRoute.name }}
+  v-toolbar.darken-3(dense, dark, color="primary")
+    v-toolbar-side-icon.logo(@click="$router.push('/')") V
+    v-breadcrumbs(:items="breadcrumbs", divider="»")
+      template(#item="props")
+        v-breadcrumbs-item(:to="props.item.to", exact) {{props.item.text}}
     v-spacer
+    //- v-btn(icon)
+    //-   v-icon {{ $vuetify.icons.settings }}
```
jeffbaumes: Please remove commented out code.

web/src/router/index.js
```diff
@@ -62,12 +55,22 @@ export const routes = [
       {
         path: 'analyze',
         name: 'Analyze Data',
```
subdavis: Remove name from wrapper route since it's used below.

web/src/App.vue
```diff
@@ -1,21 +1,49 @@
...
+    parsed = parsed.replace(`:${k}`, v);
+  });
+  return parsed;
+}
+
 export default {
   name: 'App',
+  computed: {
+    breadcrumbs() {
+      const toBreadcrumb = (route) => {
```
subdavis: prefer not to nest named functions in functions: either move to `methods` or inline as anonymous on line 43.

web/src/App.vue
```diff
@@ -1,21 +1,41 @@
...
-    v-toolbar-items
-      v-btn.px-0.square-button.darken-4(flat, color="white", to="/select")
-        v-icon {{ $vuetify.icons.close }}
-    v-spacer
-    v-toolbar-title.title
-      span.font-weight-bold.secondary--text.text--lighten-3 Biomarker Analysis »
-      span.font-weight-bold.white--text  {{ $router.currentRoute.name }}
-    v-spacer
+  v-toolbar.darken-3(dense, dark, color="primary")
+    v-toolbar-side-icon.logo(@click="$router.push('/')") V
```
subdavis: Is this a placeholder for an actual logo?  Is there an issue open for that?

web/src/App.vue
```diff
@@ -1,21 +1,41 @@
...
 <script>
+
 export default {
   name: 'App',
+  computed: {
+    breadcrumbs() {
+      return [
+        {
+          text: 'VIIME',
+          to: '/',
```
subdavis: Going forward, please prefer [named routes](https://router.vuejs.org/guide/essentials/named-routes.html) instead of absolute ones.  I'll have to go back and fix them as mentioned in #251 but

web/src/App.vue
```diff
@@ -1,21 +1,41 @@
...
+        this.toBreadcrumb(this.$route, true),
+      ];
+    },
+  },
+  methods: {
+    toBreadcrumb(route, isFull) {
+      const b = route.meta.breadcrumb;
+      return {
+        text: route.name,
+        to: route.path,
```
subdavis: same here.

web/src/router/index.js
```diff
@@ -22,10 +30,19 @@ export const routes = [
     name: 'Pretreat Data',
     component: Pretreatment,
     props: true,
+    meta: {
+      breadcrumb(params, store, isFull) {
+        const ds = store.getters.dataset(params.id);
+        return {
+          text: ds ? ds.name : params.id,
+          to: isFull ? this.path : injectParams(this.path, params),
```
subdavis: and here.

subdavis:
Not sure where nested routing ended up.  I still think the breadcrumb should match the route tree, so there probably needs to be "Analysis" in this breadcrumb: 

Screenshot from 2019-09-23 09-13-37

sgratzl:
>  I still think the breadcrumb should match the route tree, so there probably needs to be "Analysis" in this breadcrumb.

I tried and failed because of some weird Vue error. that when jumping from `analysis / anova` to `cleanup` it throws an error that `Anova...vue` has a missing `id` property

subdavis:
Not sure where nested routing ended up.  I still think the breadcrumb should match the route tree, so there probably needs to be "Analysis" in this breadcrumb: 

Screenshot from 2019-09-23 09-13-37

subdavis:
@sgratzl stray click, sorry about that.

sgratzl:
switched to named routes as suggested

subdavis:
> I tried and failed because of some weird Vue error. that when jumping from analysis / anova to cleanup it throws an error that Anova...vue has a missing id property

Hmm, let me take a look...

subdavis:
#259 may produce the desired effect.

PR #219, opened 2019-09-17
Replace D3-based ScorePlot with a C3-based one

waxlamp:
This is an initial attempt to outsource my `axisPlot` mixin to an external library.

I chose C3 because it's based on D3, which means there is theoretically a way to do arbitrary, D3-based drawing over whatever chart fundamentals it provides. My aim is to be more productive in creating classic and custom charts for efforts like Viime, while retaining the flexibility inherent in baking everything from scratch using D3. If this experiment shows a successful conversion of ScorePlot, we can replicate for the other charts and continue onward with C3 (which adding C3 to our vis utility belt).

The current state of the experiment is an attempt to use C3 to reproduce the ScorePlot component. Here's a comparison of the existing ScorePlot:
score
and the new ScorePlotC3:
c3

Note that C3 provides color legends out of the box (woo hoo!), and it comes with some nice default interactions (including a nice tooltip data display, and interactivity in the legend, which by default focuses the dataset you mouse over). Also note that I was able to shoehorn in the data ellipses.


jeffbaumes:
My opinion is divided.

Positives:
* About 50 lines less code
* The code that's there is about 50% more readable (meaning 50% less D3 code which looks foreign to standard JS devs)
* Good legend controls, including fading on legend hover

Negatives:
* Transitions are not as nice, not sure if this can be remedied. I assume it would probably require just as much complex code as straight D3.
* Has its own CSS file which might be trickier/hackier to tweak (e.g. for font consistency).
* Tooltip displays group nicely, but does not display x/y PC values in a consistent way right now.

waxlamp:
@jeffbaumes and @subdavis: this is ready for review.

waxlamp:
One question I have: should we get rid of the old ScorePlot? It has incorrect code for the ellipses, and is otherwise superseded by the C3 version.

jeffbaumes:
Yes, best to eliminate the old one in this PR. It will be there in commit history.

web/src/components/vis/ScorePlotC3.vue
```diff
@@ -0,0 +1,460 @@
...
+
+  data() {
+    return {
+      chart: null,
+      ellipseVisible: {},
+      labels: {},
+      duration: 500,
+      width: 100,
+      height: 100,
+      hasMounted: false,
```
subdavis: Maybe a short description of why this was necessary would be good to have?

subdavis:
:+1: exactly what I was thinking.  ty

PR #220, opened 2019-09-18
Rework Vuex store for new validation endpoint

subdavis:
Currently rather buggy.   This is a work in progress.


subdavis:
blocked on #212

PR #224, opened 2019-09-18
allow renaming of datasets

### LINKED ISSUE #197: Allow renaming of data sources
so far the data source has the same name as the file name (including the .csv extension). Allowing the user to rename the data source and e.g. add a description would help the user when there are a lot of data sets.

sgratzl:
closes #197

allows the user to rename a dataset and add a description to it. It stores the custom data in the `.meta` field of the DB. In addition, on the client side the `dataset.label` attribute points the the preferred label to show. The `dataset.name` is still the original file name

image

image

### DIFF
--- a/web/src/common/api.service.js
+++ b/web/src/common/api.service.js
@@ -73,6 +73,10 @@ export const CSVService = {
     return ApiService.put(`csv/${slug}/imputation`, methods);
   },
 
+  setMetaData(slug, metaData) {
+    return ApiService.put(`csv/${slug}/metadata`, metaData);
+  },
+
   updateLabel(csvSlug, changes) {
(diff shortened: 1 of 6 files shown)


jbeezley:
I think it would be best to set the `name` column in the database for this to avoid overusing the schema-less metadata.  If there is a desire to keep the original file name, we should create a column specifically for that.

subdavis:
This PR should be pretty shelf-stable and may be difficult to rebase into #131, so I'd like to get that one tidied up first before doing much more in `store/index.js`

sgratzl:
blocked by https://github.com/girder/viime/pull/233

sgratzl:
closing for now till #197 is fully specified

PR #229, opened 2019-09-18
Change docker build to install R deps in the base image

jbeezley:
@sgratzl I pushed the update you added to the base image so that the rebuild doesn't take so long.  I think it will be relatively rare that we need to change the dependency list for the R package.


jeffbaumes:
@jbeezley Since #212 is now in master, should this PR target master?

jbeezley:
Yes, rebased.

PR #234, opened 2019-09-19
Inject transformation options into GET /csv/<id>

jbeezley:
This fixes a backwards compatibility issue from #167 where the transformation, normalization, and scaling options are not returned with `GET /csv/<id>`.  This implementation is a little ham handed, so I'll probably go back and refactor a few things when there are less PR's out.

### DIFF
--- a/viime/views.py
+++ b/viime/views.py
@@ -213,8 +213,19 @@ def create_csv_file():
 
 @csv_bp.route('/csv/<uuid:csv_id>', methods=['GET'])
 def get_csv_file(csv_id):
-    csv_file = CSVFile.query.get_or_404(csv_id)
-    return jsonify(_serialize_csv_file(csv_file))
+    csv_file = _serialize_csv_file(CSVFile.query.get_or_404(csv_id))
+
+    # inject properties from the validated table model (normalization, transformation, etc.)
+    validated_table = ValidatedMetaboliteTable.query.filter_by(csv_file_id=csv_id).first()
+    if validated_table is not None:
+        transformation_schema = ValidatedMetaboliteTableSchema(
+            only=['normalization', 'normalization_argument', 'scaling',
+                  'scaling', 'transformation']
+        )
+        transformation = transformation_schema.dump(validated_table)
+        csv_file.update(transformation)
+
+    return jsonify(csv_file)
 
 
 @csv_bp.route('/csv/<uuid:csv_id>/validation', methods=['GET'])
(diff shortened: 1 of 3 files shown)


subdavis:
> a little ham handed

I don't know what that means, but I like it.

viime/views.py
```diff
@@ -213,8 +213,19 @@ def create_csv_file():
 
 @csv_bp.route('/csv/<uuid:csv_id>', methods=['GET'])
 def get_csv_file(csv_id):
-    csv_file = CSVFile.query.get_or_404(csv_id)
-    return jsonify(_serialize_csv_file(csv_file))
+    csv_file = _serialize_csv_file(CSVFile.query.get_or_404(csv_id))
+
+    # inject properties from the validated table model (normalization, transformation, etc.)
+    validated_table = ValidatedMetaboliteTable.query.filter_by(csv_file_id=csv_id).first()
+    if validated_table is not None:
```
subdavis: Perhaps I miscommunicated my issue.  I don't really need these things to come from `/csv/<id>`, but I do need them to come from `/validate`.  Right now, the response from `validate` has these properties, but they're always null regardless of what the user has set.  

The response from `POST /[normalization|transformation|scaling]` contains the correct value echoed back when I change it, but then validate is still null.

subdavis: > normalization, transformation, scaling are all missing from csv/id (expected)

I intended to say that I expected these to be missing, and that behavior was fine.

subdavis:
I think all that needs to be done is in 

```python
@csv_bp.route('/csv/<uuid:csv_id>/validate', methods=['POST'])
def save_validated_csv_file(csv_id):
    csv_file = CSVFile.query.get_or_404(csv_id)
    fatal_errors = get_fatal_index_errors(csv_file)
    if fatal_errors:
        return jsonify(validation_schema.dump(fatal_errors, many=True)), 400

    old_table = ValidatedMetaboliteTable.query.filter_by(csv_file_id=csv_id)
    try:
        if old_table is not None:
            old_table.delete()
            db.session.commit()  # we actually want to persist to invalidate the old table

        validated_table = ValidatedMetaboliteTable.create_from_csv_file(csv_file)
        db.session.add(validated_table)
        db.session.commit()
        return jsonify(validated_metabolite_table_schema.dump(validated_table)), 201
    except Exception:
        db.session.rollback()
        raise
```

Before the old validation is deleted, could we copy the values of the transforms into the new validation?

jbeezley:
Yes, that is the issue.  I did that intentionally to ensure that there are no errors in the transformation pipeline.  For example, the user could mask a metadata column that was used for normalization. 

How do you think this should be handled?  Try to keep the same transformation attributes, but fall back to resetting if there is an error?

jbeezley:
I think there are user experience issues with the transformations getting reset that we will want to address.  It may involve asking users what is the "least surprising" outcome.

PR #236, opened 2019-09-19
simple scale buttons for plots

### LINKED ISSUE #202: Transform Table Plot Sizes
with the newly added scree plot (#182). The Transform Table looks like this on a regular Full HD screen: 

image

possible ways to deal with that:
 * allow to resize the plots in a granular way (e.g. small, medium, large). a  CSS grid layout could be used for layouting
 * allow to en/disable individual plots
 * auto adapt the plots and plot sizes to the available screen (disadvantage: unbalanced aspect ratios)

sgratzl:
closes #202 by adding scale buttons to plots and option to disable the plot

image

image

### DIFF
--- a/web/src/components/LayoutGrid.vue
+++ b/web/src/components/LayoutGrid.vue
@@ -1,0 +1,41 @@
+<script>
+export default {
+  props: {
+    cellSize: {
+      type: Number,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      gridHint: 4,
+    };
+  },
+  computed: {
+    gridStyle() {
+      return {
+        gridTemplateColumns: `repeat(${this.gridHint}, ${this.cellSize}px)`,
+        gridAutoColumns: `${this.cellSize}px`,
+        gridAutoRows: `${this.cellSize}px`,
+      };
+    },
+  },
+  mounted() {
+    const available = this.$el.getBoundingClientRect().width;
+    // compute the number of horizontal cells based on the container width
+    this.gridHint = Math.floor(available / this.cellSize);
+  },
+};
+</script>
+
+<template lang="pug">
+.gridlayout(:style="gridStyle")
+  slot
+</template>
+
+<style scoped lang="scss">
+.gridlayout {
+  display: grid;
+  grid-gap: 0.5em;
+}
+</style>
--- a/web/src/components/Transform.vue
+++ b/web/src/components/Transform.vue
@@ -9,12 +9,14 @@ import {
 import ScorePlotTile from '@/components/vis/ScorePlotTile.vue';
 import ScreePlotTile from '@/components/vis/ScreePlotTile.vue';
 import LoadingsPlotTile from '@/components/vis/LoadingsPlotTile.vue';
+import LayoutGrid from './LayoutGrid.vue';
 
 export default {
   components: {
     LoadingsPlotTile,
     ScorePlotTile,
     ScreePlotTile,
+    LayoutGrid,
   },
   props: {
     id: {
@@ -32,6 +34,8 @@ export default {
         scree: true,
         loadings: true,
       },
+      cellSize: 300,
+      plotHeader: 48 + 5,
     };
   },
   computed: {
@@ -44,6 +48,8 @@ export default {
     norm_arg() { return this.$store.getters.txType(this.id, 'normalization_argument'); },
     trans() { return this.$store.getters.txType(this.id, 'transformation'); },
     scaling() { return this.$store.getters.txType(this.id, 'scaling'); },
+    plotWidth() { return this.cellSize * 2; },
+    plotHeight() { return this.cellSize * 2 - this.plotHeader; },
   },
   methods: {
     async transformTable(value, category, argument, methods) {
@@ -128,21 +134,21 @@ v-layout.transform-component(row, fill-height)
     h4.display-1.pa-3 Loading Data Set
   v-layout(column, v-else-if="ready && valid")
     v-container.grow-overflow.ma-0(grid-list-lg, fluid)
-      v-layout(row, wrap)
+      layout-grid(:cell-size="cellSize")
         score-plot-tile(
-            v-if="visiblePlots.score", :key="score",
-            :width="600",
-            :height="600",
+            v-if="visiblePlots.score", key="score",
+            :width="plotWidth",
+            :height="plotHeight",
             :id="id")
         loadings-plot-tile(
-            v-if="visiblePlots.loadings", :key="loadings",
-            :width="600",
-            :height="600",
+            v-if="visiblePlots.loadings", key="loadings",
+            :width="plotWidth",
+            :height="plotHeight",
             :id="id")
         scree-plot-tile(
-            v-if="visiblePlots.scree", :key="scree",
-            :width="600",
-            :height="600",
+            v-if="visiblePlots.scree", key="scree",
+            :width="plotWidth",
+            :height="plotHeight",
             :id="id")
   v-container(v-else-if="ready", fill-height)
     v-layout(column)
@@ -158,4 +164,5 @@ v-layout.transform-component(row, fill-height)
 .checkboxlist {
   display: block;
 }
+
 </style>
--- a/web/src/components/Transform.vue
+++ b/web/src/components/Transform.vue
@@ -136,18 +136,18 @@ v-layout.transform-component(row, fill-height)
     v-container.grow-overflow.ma-0(grid-list-lg, fluid)
       layout-grid(:cell-size="cellSize")
         score-plot-tile(
-            v-if="visiblePlots.score", key="score",
+            v-show="visiblePlots.score",
             :width="plotWidth",
             :height="plotHeight",
             :columns="dataset.column.data",
             :id="id")
         loadings-plot-tile(
-            v-if="visiblePlots.loadings", key="loadings",
+            v-show="visiblePlots.loadings",
             :width="plotWidth",
             :height="plotHeight",
             :id="id")
         scree-plot-tile(
-            v-if="visiblePlots.scree", key="scree",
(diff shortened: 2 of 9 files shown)


subdavis:
Neat, I think something like this would be handly.

I had a similar idea, but it involved swapping between `vis-tile` and `vis-tile-large` such that there were only 2 sizes of plot: default and full width.  The nice part about that would be, with the roomier `vis-tile-large`, the menu controls could automatically jump into a left nav pane.

Regardless of implementation, I think providing more than those two options may be overkill, but I would like to hear your opinion.

subdavis:
You could imagine something a bit like this, but with the plot the appropriate height and width.

Screenshot from 2019-09-19 10-40-43

sgratzl:
@subdavis can you please your comments to #202. The issue is a better place to discuss how the issue should be implemented

sgratzl:
grid compacting (works best with more plots)

image

sgratzl:
image

image


when one disables and enables the plot, the plot won't be rerended. @waxlamp do you have an idea why?

subdavis:
@sgratzl it's because of a bug @waxlamp and I talked about in #195 where the plot doesn't render at mount.

I think it should be fixed on master.  Is it possible to rebase and try again?

subdavis:
You could also sidestep the issue by using v-show instead of v-if for now and fix it later.

sgratzl:
extracted bug to https://github.com/girder/viime/issues/244

web/src/components/LayoutGrid.vue
```diff
@@ -0,0 +1,41 @@
...
+.gridlayout(:style="gridStyle")
+  slot
+</template>
+
+<style scoped lang="scss">
+.gridlayout {
+  display: grid;
+  grid-gap: 0.5em;
+}
+</style>
```
subdavis: +1  for making this a separate component.

web/src/components/Transform.vue
```diff
@@ -27,6 +29,13 @@ export default {
       normalize_methods,
       transform_methods,
       scaling_methods,
+      visiblePlots: {
+        score: true,
+        scree: true,
+        loadings: true,
+      },
+      cellSize: 300,
+      plotHeader: 48 + 5,
```
subdavis: Could you add a comment about where 48 and 5 came from?  +1 for separating them out an not just writing `53`.

web/src/components/vis/LoadingsPlotTile.vue
```diff
@@ -51,13 +51,14 @@ export default {
...
 <template lang="pug">
 vis-tile(v-if="plot", title="PCA Loadings Plot", :loading="plot.loading")
-  loadings-plot(
-      :width="width",
-      :height="height",
-      :points="plot.data",
-      :pc-x="pcX",
-      :pc-y="pcY",
-      :show-crosshairs="showCrosshairs")
+  template(#default="wrapper")
```
subdavis: Is this change relevant to this PR?   It's fine if it isn't, just curious.

web/src/components/vis/VisTile.vue
```diff
@@ -11,28 +11,61 @@ export default {
...
+    };
+  },
+
   computed: {
-    hasControls() {
-      return !!this.$slots.controls;
+    smallEnough() {
+      return this.scale <= 0.5;
+    },
+    largeEnough() {
```
subdavis: Prefer better names.  perhaps `isMaxScale` and `isMinScale`.

web/src/components/vis/VisTile.vue
```diff
@@ -11,28 +11,61 @@ export default {
...
+      return this.scale >= 4;
+    },
+    scaleClass() {
+      return `scale${Math.round(this.scale * 100)}`;
```
subdavis: `* 100` seems unnecessary.  Why not just have your min and max be 50 and 400 respectively?

Actually, why was multiplication used here? if the point is to have N different classes which apply N different grid sizes, why not just use addition?  `scale += 1` and `scale -= 1` would achieve the same thing without the risk of floating point issues.  You're already stuck on an exponential scale, so it's not like you could add scale=500 if you wanted something in the middle anyway.

web/src/components/vis/VisTile.vue
```diff
@@ -11,28 +11,61 @@ export default {
...
+  },
+
   computed: {
-    hasControls() {
-      return !!this.$slots.controls;
+    smallEnough() {
+      return this.scale <= 0.5;
+    },
+    largeEnough() {
+      return this.scale >= 4;
```
subdavis: Perhaps refactor `minScale` and `maxScale` to constants at the top of the script.

web/src/components/vis/VisTile.vue
```diff
@@ -11,28 +11,61 @@ export default {
...
+}
+.scale100 {
+  grid-column: span 2;
+  grid-row: span 2;
+}
+.scale200 {
+  grid-column: span 4;
+  grid-row: span 4;
+}
+.scale400 {
```
subdavis: prefer `scale1, scale2, scale3` etc.  The actual numbers don't seem to matter, so you could use grid layouts that aren't powers of 2.

web/src/components/vis/LoadingsPlotTile.vue
sgratzl: I'm accessing the `wrapper.scale` one line below and I don't know another way to assign a scoped prop

web/src/components/vis/VisTile.vue
sgratzl: those methods just exists because of limitation of pug

web/src/components/vis/VisTile.vue
sgratzl: `span 4` refers to that four cells should be used. `scale200` mean a scaling factor of  200 percent. Thus, you can read it as: When the scaling factor is `200`% the cell should span over four cells in the grid layout

subdavis: I suppose my point is that the math here, while technically correct, isn't necessary.  You've already hard-coded the result of the equation `cell width * 200%` within the CSS class.

All you really have is a finite set of constant classes you can apply, and you iterate through them by generating their string keys mathematically. 

I think you could reduce the complexity by having something like below where you just have buttons that increment a counter up and down.  A bonus here would be that your min and max go away, and you iterate from 0 to `sizeClasses.length - 1`

```
<script>
  ...
  data() { return { sizeClasses = [ 'class1', 'class2', 'class3', ...] }},
<template>
   ...
   <element :class="sizeClasses[index]"
</template>
```

web/src/components/vis/VisTile.vue
subdavis: This discussion has moved to the thread below.

web/src/components/vis/VisTile.vue
subdavis: Move this conversation to the thread below.

web/src/components/vis/VisTile.vue
subdavis: Another consideration is that the math ties us unnecessarily to this exponential scale.  We might find that a Fibonacci scale is more appropriate, and if the class name generation weren't mathematical, all you have to do is change a few CSS properties.

waxlamp: +1 for simplifying the logic here.

web/src/components/vis/VisTile.vue
```diff
@@ -11,28 +11,61 @@ export default {
...
   },
 
+  data() {
+    return {
+      scale: 1,
+    };
+  },
+
   computed: {
-    hasControls() {
```
waxlamp: It doesn't seem like this should have been deleted--it's meant to control whether the menu icon appears in the toolbar.

sgratzl: no it controlled whether the overall toolbar on top was rendered. The menu icon is rendered in the Plot tiles directly.

sgratzl:
changed version:

image

sgratzl:
feedback is integrated

web/src/components/vis/VisTile.vue
```diff
@@ -11,28 +13,91 @@ export default {
...
     v-toolbar.primary.darken-3.top-rounded(dark, flat, dense)
       v-toolbar-title {{ title }}
       v-spacer
-      v-toolbar-items(v-if="hasControls")
+      v-toolbar-items
+        v-btn(@click="scaleIndex--", :disabled="scaleIndex === 0", icon)
+          v-icon {{ $vuetify.icons.magnifyMinus }}
+        v-select.scaleFactor(:value="scaleIndex", @change="scaleIndex = parseInt($event, 10)",
+            :items="scaleOptions", hide-details, item-text="text", item-value="value")
+        v-btn(@click="scaleIndex++", :disabled="scaleIndex === scales.length - 1", icon)
```
subdavis: Better not to rely on the Vue event loop to protect you from invalid states.  If the page or the machine is under heavy load, `:disabled` may not compute fast enough. 

Almost every project I've got right now has [this](https://stackoverflow.com/questions/11409895/whats-the-most-elegant-way-to-cap-a-number-to-a-segment) in the utilities.

Recommend:

```
script
 import { clamp } from 'utils';
  methods
    setScaleIndex(value) { this.scaleIndex = clamp(value, min, max); },
...
template
  v-btn(@click="setScaleIndex(scaleIndex +/- 1)", ...)
```

subdavis: To be clear, I don't use the Number prototype, I just take the number as the first argument.

web/src/components/Transform.vue
sgratzl: can be removed when https://github.com/girder/viime/issues/256 is implemented

web/src/components/vis/VisTile.vue
sgratzl: done

PR #233, opened 2019-09-19
Resolve validation bugs

subdavis:
closes #215

PR #237, opened 2019-09-19
Wilcoxon Refactoring

### LINKED ISSUE #232: Update Wilcoxon test to output p-value table of metabolite changes for each pair of groups, and heatmap of z-scores of low-p-value metabolites

sgratzl:
closes #232 

based on #233

image

notes
 * rebuild your docker container

### DIFF
--- a/web/src/components/vis/AnovaTable.vue
+++ b/web/src/components/vis/AnovaTable.vue
@@ -1,5 +1,4 @@
 <script>
-import { format } from 'd3-format';
 import MetaboliteTable from './MetaboliteTable.vue';
 
 export default {
--- a/web/src/components/vis/AnovaTable.vue
+++ b/web/src/components/vis/AnovaTable.vue
@@ -1,10 +1,16 @@
 <script>
 import { format } from 'd3-format';
+import MetaboliteTable from './MetaboliteTable.vue';
 
 export default {
 
+  components: {
+    MetaboliteTable,
+  },
+
   props: {
     data: {
+      type: Object,
       validator: prop => !prop || ('data' in prop && 'pairs' in prop),
       required: true,
     },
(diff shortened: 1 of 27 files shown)


web/src/components/vis/AnovaTable.vue
```diff
@@ -32,33 +31,23 @@ export default {
         },
         {
           text: 'Intercept',
-          align: 'right',
```
subdavis: Why remove this alignment?  Column headers should be aligned with data.

sgratzl: I don't know any spreadsheet application that automatically aligns column labels to the right if the values are numbers

subdavis: It only caught my eye because headers are closer to their neighbor column than their own.  The ui here also doesn't really make me think "spreadsheet" like the cleanup table does.

For what it's worth, I believe spreadsheets align numbers to the right to be consistent with accounting software and books, so that convention may or may not be important to a user (I don't really know).

sgratzl: well they do it cause it easier to compare numbers that way:
```
23123.23
  232.23
```

vs

```
23123.23
232.23
```

subdavis: No, it's not a blocker.

web/src/components/vis/analyses.js
```diff
@@ -1,28 +1,24 @@
...
+import { plot_types } from '../../utils/constants';
 
 export default [
   {
     path: 'wilcoxon',
-    name: 'Wilcoxon signed-rank test',
+    name: 'Wilcoxon test',
     shortName: 'Wilcoxon Test',
-    description: 'The Wilcoxon signed-rank test tests the null hypothesis that two related paired samples come from the same distribution. In particular, it tests whether the distribution of the differences x - y is symmetric about zero. It is a non-parametric version of the paired T-test.',
+    description: 'TODO R custom code',
```
subdavis: Should this TODO be changed?

sgratzl: well the only one who can write a proper text are our collaborators

PR #238, opened 2019-09-19
move imputation settings to side panel

### LINKED ISSUE #200: Put imputation settings in the sidebar somewhere
the clean up table uses a dialog to set the imputation settings. 

image

the transform table (and analysis) use a left side panel to change options

image

So far, the imputation settings are hidden and cannot be easily explored.

sgratzl:
closes #200

image

### DIFF
--- a/web/src/components/Cleanup.vue
+++ b/web/src/components/Cleanup.vue
@@ -34,10 +34,8 @@ export default {
       },
       defaultColOption,
       defaultRowOption,
-      settingsDialog: false,
       mnarImputationMethods: mnar_imputation_methods,
       mcarImputationMethods: mcar_imputation_methods,
-      imputation: {},
     };
   },
   computed: {
@@ -65,11 +63,11 @@ export default {
       }
       return `Column ${base26Converter(members[0] + 1)}`;
     },
-  },
-  watch: {
-    dataset() {
-      this.imputation.mnar = this.dataset.imputationMNAR;
-      this.imputation.mcar = this.dataset.imputationMCAR;
+    mnar() {
+      return this.dataset.imputationMNAR;
+    },
+    mcar() {
+      return this.dataset.imputationMCAR;
     },
   },
   methods: {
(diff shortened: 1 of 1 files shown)


subdavis:
I don't think we should dedicate another full vertical column for something like the imputation settings. In general, I think this features them too prominently proportional to how important they are to a user.

I'm especially hesitant because there will be 3 columns on the left if any validation errors are present.   The validation errors are much more important than imputation, and putting this extra column between the validation error list and the body of the table may confuse users, e.g. "Am I supposed to use this to resolve validation errors in some way"?

I don't know what the right place to put it is.  Perhaps we could float this column to the right side of the table and collapse it some way.

jeffbaumes:
From my comment in #200:

> One solution is to have a new Imputation tab between Clean Up and Transform, and the options should appear in the second sidebar, with a view of the table with imputed values filled in.

subdavis:
Putting a new Imputation tab after cleanup could potentially cause some corner cases where you clean up your dataset then move on to imputation and change a setting that changes the variance among a group and creates new validation warnings.  

@jeffbaumes @sgratzl how do you feel about imputation as a permanent sub-route underneath cleanup instead of adjacent to it?  So it would be at the same level as warnings and errors.

I think I agree with @sgratzl's goal of keeping imputation somewhere within the cleanup route, I just don't know if a full second sidebar is the best place to put it unless it were able to be hidden until called upon.

jeffbaumes:
> Putting a new Imputation tab after cleanup could potentially cause some corner cases where you clean up your dataset then move on to imputation and change a setting that changes the variance among a group and creates new validation warnings.

I see. So the process must be something like:

1. Look for missingness validation errors
2. Missing data is filled in
3. Look for variance validation errors

This intertwines the two and makes them need to be in the same "step", unless we want to split into 3 steps which would be overkill.

> @jeffbaumes @sgratzl how do you feel about imputation as a permanent sub-route underneath cleanup instead of adjacent to it? So it would be at the same level as warnings and errors.

Seems reasonable. I can't think of a better idea right now.

sgratzl:
closed in favor of #245

PR #246, opened 2019-09-20
Add validation errors for rows/column containing NaN's

jbeezley:
This promotes data rows/columns containing over 95% NaN's (or non-numeric values) to a validation error rather than a warning.  I'm not sure how the imputation code handles this case, but clearly it indicates a problem that should block further processing.

(The 95% value is arbitrary, maybe something lower or higher is more appropriate.  I mostly wanted to handle cases where a column is actually categorical metadata.)


jeffbaumes:
@jbeezley good to merge?

PR #253, opened 2019-09-23
[WIP] Landing Page

### LINKED ISSUE #243: Front page
I've made a rough [mockup](https://docs.google.com/drawings/d/1XFh0NTVz_ctDip2LA9C5jcvVIIieYPuh5rnoOCKbwH4/edit?usp=sharing). @jtomeck could you put this on your list to directly code up in Vue? This page would get the root route in the app, perhaps @subdavis could get you set up with an initial empty .vue file.

The following issues can be solved as part of this:
- #309 
- #310

subdavis:
closes #243


jtomeck:
Please feel free to rip the code apart. It's quite messy due to my lack of experience with pug and there is much I was uncertain of how to do.

**To do:**
1. Footer with disclaimer / licensing info?
2. Fill in button links where necessary
3. Plan the functionality for where I've added `<join us link here>`
4. Create favicon
5. Discuss adding VIIME logo to the application toolbar

Let me know if I've missed anything in this list.

P.S. - @subdavis I was not able to request you as a reviewer, so I'm mentioning you

subdavis:
For anyone looking at this, the actual app will be completely broken.  I will try and resolve the conflicts with master asap and try not to step on Jared's toes.

sgratzl:
> * Create favicon
> * Discuss adding VIIME logo to the application toolbar

I would create separate PRs for that since they are independent of the landing page

subdavis:
@jtomeck I've resolved the changes on another branch: https://github.com/girder/viime/tree/landing-page-fix

I'd like to force update this branch to be level with that one.  It was easier to resolve this by starting with current master, writing the router changes again since so much has changed, then cherry picking your changes over top of that.

None of what you've done in Landing.vue will change.  Is that OK?

jeffbaumes:
Good point @sgratzl. The app bar change in particular might need some iteration so I can see that one being done in another PR.

jeffbaumes:
Thanks @subdavis for updating the branch. I can get it working now. The page is responsive when thin, but does not look good wide, since the main text goes up into the logo:

image

jtomeck:
> Thanks @subdavis for updating the branch. I can get it working now. The page is responsive when thin, but does not look good wide, since the main text goes up into the logo:

Looks like it might need a height breakpoint as well to prevent that overlap.

subdavis:
Hmm, when it's thin I see stacked scrollbars in chrome. Same in firefox. Happens around the 915px width.

Screenshot from 2019-10-24 15-42-41

jeffbaumes:
I may have spoken too soon on it working thin. I am not getting the same effect as when you showed it at the meeting this morning. It seems to not compress well thinner than 900px.

waxlamp:
Are we looking to do design feedback on this PR, or just code quality stuff?

Also, I sort of assumed that the landing page was going to be in an independent repository, and be deployed to production as a static site alongside Viime, rather than as part of the Flask app itself, if that makes sense. Is there an advantage to doing it this way?

jeffbaumes:
> Also, I sort of assumed that the landing page was going to be in an independent repository, and be deployed to production as a static site alongside Viime, rather than as part of the Flask app itself, if that makes sense. Is there an advantage to doing it this way?

Good point. I had similar thoughts and I discussed this a bit with @jtomeck yesterday, that we could have gone that other route. We could certainly do so in the future. I think we're exploring some of the tradeoffs right here with this PR. A separate repo would complicate deployment a bit, and might disallow/complicate some potential tie-ins like having a contact us mechanism on the front page that functions the same as a feedback link in the app. A separate marketing front page has the benefits of letting Jared not struggle with Vue/Vuetify and focus on HTML/CSS. Jared said there were some struggles, but that it was not overwhelmingly unproductive working in the "dev" environment.

jeffbaumes:
Might as well also mention that the structure of this PR (and the fact that it's been intended to be included in this repo) has been around for a month, so a different method could have been proposed earlier. I'd like to commit to this route for viime at this point.

subdavis:
What would the goal of such a separation be?  There are a couple major drawbacks to the current approach.

1. Not possible for search engines to index the page. 
2. Allows users without JavaScript enabled to see what we're about
3. Makes development easier for Jared (maybe)

I see 1 as by far the most important (sorry Jared) and Nuxt or something like it could help us with 1 and 2.

waxlamp:
> Might as well also mention that the structure of this PR (and the fact that it's been intended to be included in this repo) has been around for a month, so a different method could have been proposed earlier. I'd like to commit to this route for viime at this point.

This, combined with the observation that we can always split the static site out later if we want, seems to be the governing principle we should follow for now.

@subdavis's observation that search engines won't index the homepage is super important, but if Nuxt can help with that, then it seems like a good solution for now (though, I will admit to trying to learn what Nuxt is by reading its homepage, which seems determined not to explain it :joy_cat: ).

One last thought on this topic: if viime.org went to a static webpage, and the app lived at app.viime.org, that would be pretty cool. I don't know why I feel this way but that setup feels professional and solid for some reason.

jtomeck:
> I may have spoken too soon on it working thin. I am not getting the same effect as when you showed it at the meeting this morning. It seems to not compress well thinner than 900px.

I found that there's a `min-width: 900px;` on `#app` that I don't believe was part of my changes, which explains why it wouldn't compress any thinner. I'm assuming this has been added in order to keep the tables legible width-wise in the app (I suppose another downside of having the landing page live within the app).

@subdavis - do you have any suggested course of action for this? I'm sure it's been added for a reason, so I don't want to remove it. Maybe add a `.landing` class to it and add `min-width: auto;` to that class?

sgratzl:
re viime mask with the screenshot in the background: we can think of using: https://developer.mozilla.org/de/docs/Web/CSS/mask to do the masking on the fly. This should simplify switching images or animating the background.

jtomeck:
> re viime mask with the screenshot in the background: we can think of using: https://developer.mozilla.org/de/docs/Web/CSS/mask to do the masking on the fly. This should simplify switching images or animating the background.

Yeah, I'm still debating whether we even want that image for the banner area. I think I'd really like a motion gif showing someone interacting with the plots or something similar. If we stick with this, I think this is a good idea. I've never used CSS masking before, so this is neat.

jtomeck:
I believe with these commits I'm ready for everyone to review things. Specifically the following:

- What is the best way to handle removing the `min-width: 900px;` on `#app`, but only on the landing page? (This is the reason for responsive styles not working on the landing page)
- What should the form fields be under Collaboration? (this form will also need functionality added)
- What should I add under license in the footer?

Besides these things, I would like to figure out imagery for the rest of the placeholders down the page. If anyone has time to actually review the code as well, that would be valuable to me since I am inexperienced with pug and am not sure I'm doing everything correctly.

subdavis:
Minor fixes in #420

sgratzl:
>     * What is the best way to handle removing the `min-width: 900px;` on `#app`, but only on the landing page? (This is the reason for responsive styles not working on the landing page)

should be fixed with https://github.com/girder/viime/pull/253/commits/cdf4d5dcbfa1ff8b08a20339b1ffb5a967b0caf5

subdavis:
I'm going to follow Sam's lead here and say I think the images, footer, and form fields can all be dealt with separately.  Would that be OK, @jtomeck ?

jeffbaumes:
For the footer, let's follow data.kitware.com and put a Privacy Notice link to https://www.kitware.com/privacy

jeffbaumes:
But let's merge and fix all these things after the fact.

PR #255, opened 2019-09-23
Correlation network analysis

### LINKED ISSUE #249: Correlation network analysis
Compute the correlation matrix between metabolites, and layout a network with edges where the absolute value of correlation is above some threshold, which should be configurable. Should look something like this:

image

sgratzl:
closes #249 

image

Notes
 * please run `yarn` to install d3-force
 * node coloring is not possible yet, since there are no groupings for columns yet


sgratzl:
>     5\. Encoding a value (not sure which one) on the node color would also be cool.

as mentioned in the PR comment. I didn't know a value either. When we have merged data sources we have the original source as categorical attribute

waxlamp:
We could also use a continuous variable for the colormap as well.

sgratzl:
> * When the link distance is changed, the links stop working (they stay in their last valid place, and the nodes update properly, but without any visible links between them).

fixed

> 
> * The link distance shouldn't be clamped to 100.

fixed

> 
> * The links seem too thick to me. It seems like you're encoding the correlation value on them, but perhaps a more nonlinear scale that makes only the most correlated links significantly thicker would be good.

switched to power2 scale where the min correlation will be mapped to 1px up to 1 = 8px

image

web/src/components/vis/CorrelationTile.vue
```diff
@@ -0,0 +1,92 @@
...
+              hide-details, min="0", max="1", step="0.01",
+              @change="changePlotArgs({min_correlation: $event})")
+
+    v-toolbar.darken-3(color="primary", dark, flat, dense, :card="false")
+      v-toolbar-title Advanced Options
+    v-card.mx-3(flat)
+      v-card-actions
+        v-layout(column)
+          v-switch(v-model="showLabels", label="Node Labels")
+          v-text-field(v-model="linkDistance", label="Link Distance",
```
subdavis: Please add `hide-details` to all `v-input` derivatives if you aren't using the feature.  It makes everything look more compact.

web/src/components/vis/CorrelationTile.vue
```diff
@@ -0,0 +1,92 @@
...
+          v-slider.minCorrelation(:value="min_correlation", label="0", thumb-label,
+              hide-details, min="0", max="1", step="0.01",
+              @change="changePlotArgs({min_correlation: $event})")
+
+    v-toolbar.darken-3(color="primary", dark, flat, dense, :card="false")
+      v-toolbar-title Advanced Options
+    v-card.mx-3(flat)
+      v-card-actions
+        v-layout(column)
+          v-switch(v-model="showLabels", label="Node Labels")
```
subdavis: same here.

web/src/components/vis/CorrelationTile.vue
```diff
@@ -0,0 +1,92 @@
...
+  template(#controls)
+    toolbar-option(title="Method", :value="plot.args.method",
+        :options="correlation_methods",
+        @change="changePlotArgs({method: $event})")
+    v-toolbar.darken-3(color="primary", dark, flat, dense, :card="false")
+      v-toolbar-title Minimum Correlation
+    v-card.mx-3(flat)
+      v-card-actions
+        v-layout(column)
+          v-slider.minCorrelation(:value="min_correlation", label="0", thumb-label,
```
subdavis: same here

PR #259, opened 2019-09-24
Subdavis/appbar

subdavis:
This seems to have the desired effect.  Will this work?

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -25,6 +25,9 @@ export default {
           .map(route => ({
             text: route.name || route.path,
             to: { name: route.name, params: this.$route.params },
+            ...(typeof route.meta.breadcrumb === 'function'
+              ? route.meta.breadcrumb.call(route, this.$route.params, this.$store)
+              : {}),
           })),
       ];
     },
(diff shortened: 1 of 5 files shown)


sgratzl:
I ran in the same error

please do the following to reproduce the error
 1. open ANOVA
 2. click on Clean Up Table.

I get the following error:
```
vue.esm.js?a026:628 [Vue warn]: Missing required prop: "id"

found in

---> <AnovaTableTile> at src/components/vis/AnovaTableTile.vue
       <RouteWrapper> at src/components/RouteWrapper.vue
         <Pretreatment> at src/components/Pretreatment.vue
           <VApp>
             <App> at src/App.vue
               <Root>
...
vue.esm.js?a026:628 [Vue warn]: Error in callback for watcher "plot.valid": "TypeError: Cannot read property 'valid' of undefined"
```

subdavis:
Good catch, I  Can't tell if this is a bug introduced by the code or an existing bug we've just uncovered.  I'll see if I can figure it out for a quick minute and if not we can just merge what you have.

sgratzl:
I didn't find the thing within an hour. In the end I think the reason is the `keep-alive` from `Pretreatment.vue`. Tho, I have no idea why this is an issue right now

PR #260, opened 2019-09-24
transformation boxplot

sgratzl:
closes #193 and closes #248 

image

image

Issue #208, opened 2019-09-17
Data Parsing


sgratzl:
right now the `/csv/<id>` endpoint of a small dataset returns:

```js
{
  "columns": [
    {
      "column_header": "Name",
      "column_index": 0,
      "column_type": "key"
    },
    {
      "column_header": "m1",
      "column_index": 1,
      "column_type": "group"
    },
    {
      "column_header": "m2",
      "column_index": 2,
      "column_type": "measurement"
    },
    {
      "column_header": "m3",
      "column_index": 3,
      "column_type": "measurement"
    }
  ],
  "created": "2019-09-16T19:50:16.045404+00:00",
  "groups": "Name,m1\nA,1\nB,2\nC,3\nD,1\nE,32\nF,3\n",
  "id": "ce366700-af03-4bec-a4ed-3eda4ad28c6e",
  "imputation_mcar": "random-forest",
  "imputation_mnar": "zero",
  "measurement_metadata": "Name,m2,m3\n",
  "measurement_table": ",m2,m3\nA,5,4\nB,3,3\nC,4,3\nD,34,4\nE,3,2\nF,4,4\n",
  "meta": {},
  "name": "test-Sheet2.csv",
  "rows": [
    {
      "row_index": 0,
      "row_name": "Name",
      "row_type": "header"
    },
    {
      "row_index": 1,
      "row_name": "A",
      "row_type": "sample"
    },
    {
      "row_index": 2,
      "row_name": "B",
      "row_type": "sample"
    },
    {
      "row_index": 3,
      "row_name": "C",
      "row_type": "sample"
    },
    {
      "row_index": 4,
      "row_name": "D",
      "row_type": "sample"
    },
    {
      "row_index": 5,
      "row_name": "E",
      "row_type": "sample"
    },
    {
      "row_index": 6,
      "row_name": "F",
      "row_type": "sample"
    }
  ],
  "sample_metadata": "Name\nA\nB\nC\nD\nE\nF\n",
  "size": 64,
  "table": "Name,m1,m2,m3\nA,1,5,4\nB,2,3,3\nC,3,4,3\nD,1,34,4\nE,32,3,2\nF,3,4,4\n",
  "table_validation": []
```

[...]

PR #265, opened 2019-09-25
data parsing

sgratzl:
based on #260 

closing #208 

changed the data transfer and try to avoid unneeded transfers

* `/<id>` -> removed `groups, measurement_metadata, sample_metadata`
* `/<id>` -> transfer `measurement_table` = imputed table as JSON using `to_dict(orient=split)`
* `/<id>/validate` -> transfer `measurements, groups, sample_metadata, measurement_metadata` as JSON. However, I drop index/column when not needed and replace them with the columns/index from the `/<id>` `measurement_table` since the information is redundant otherwise.

client new attributes in a dataset (`this.dataset`) all having the following structure

```ts
interface IClientDataFrame {
 columnNames: string[];
 rowNames: string[];
 data: (number | string)[][]; // cell = data[rowIndex][columnIndex]
}
```
 * `measurement_table` ... imputed data table
 * `validatedMeasurements`
 * `validatedGroups`
 * `validatedMeasurementsMetaData`
 * `validatedSampleMetaData`


viime/models.py
```diff
@@ -311,18 +309,13 @@ def fix_file_name(self, data, **kwargs):
...
             return data
 
         data['table'] = data['table'].to_csv(header=False, index=False)
-        data['measurement_table'] = get_csv('measurement_table')
-        data['measurement_metadata'] = get_csv('measurement_metadata')
-        data['sample_metadata'] = get_csv('sample_metadata')
-        data['groups'] = get_csv('groups')
+        if data.get('measurement_table') is not None:
+            data['measurement_table'] = data['measurement_table'].to_dict(
+                orient='split')
```
jbeezley: Any reason you aren't converting the main table as well?

sgratzl: cause in the cleanup table the source table is many treated as "raw" file (e.g. no meaning of column headers, row headers, typing, ...). Thus, I wanted to keep it similar without converting

web/src/store/index.js
```diff
@@ -171,18 +172,29 @@ const mutations = {
...
         scaling: data.scaling,
+
+        // imputed measurements
+        measurement_table,
+
+        // reset
+        validatedMeasurements: null,
+        validatedGroups: null,
+        validatedMeasurementsMetaData: null,
+        validatedSampleMetaData: null,
```
subdavis: Same comment as in #260 : please include these in the defaults and include `measurement_table` as well.

I'll just approve this so you can do it on either branch.

PR #266, opened 2019-09-26
clustered heatmap

### LINKED ISSUE #267: create clustered heatmap

sgratzl:
closes #267 

based on #265 (validated data in store) and #255 (vue-resize)

first version of a hierarchical clustered heatmap of the validated measurements. 

image

From an interaction point of view, so far only highlighting of rows/columns. Collapsing a level requires some kind of aggregation (median?, mean?)

### DIFF
--- a/viime/analyses.py
+++ b/viime/analyses.py
@@ -1,6 +1,8 @@
 from typing import Any, Dict
 
+import numpy as np
 import pandas as pd
+from scipy.cluster.hierarchy import linkage
 
 from .opencpu import opencpu_request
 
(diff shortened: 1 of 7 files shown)


jeffbaumes:
I think mean would be fine for aggregation.

waxlamp:
Is it possible to use a more ["standard"](https://en.wikipedia.org/wiki/Dendrogram#/media/File:Heatmap_RNAseqV2_1.png) dendrogram? Or do we like the wavier style?

jeffbaumes:
I'd think it would be an easy change (or can be made an option). We can ask Tom which he prefers.

sgratzl:
image

image

sgratzl:
image

subdavis:
pending @jbeezley  I think?

PR #268, opened 2019-09-26
verify file extensions

### LINKED ISSUE #199: File extension validation
atm the user can select any file to upload, even images. However, so far only CSV and soon Excel files are supported. Allowing only valid data types e.g. by setting the `accept` flag (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#Unique_file_type_specifiers) will avoid those kind of errors

image

sgratzl:
closes #199

### DIFF
--- a/web/src/components/Upload.vue
+++ b/web/src/components/Upload.vue
@@ -69,19 +69,27 @@ export default {
   },
   methods: {
     async onFileChange(targetFiles) {
-      this.pendingFiles = this.pendingFiles.concat([...targetFiles].map(file => ({
-        file,
-        status: 'pending',
-        progress: {},
-        meta: {},
-      })));
-
       const excelMimeTypes = [
         'application/vnd.ms-excel',
         'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
       ];
 
       const isExcelFile = file => excelMimeTypes.includes(file.type) || file.name.match(/\.xlsx?$/i);
+      const isCSVFile = file => file.type === 'text/csv' || file.name.match(/\.csv$/i);
+      const isTextFile = file => file.type === 'text/plain' || file.name.match(/\.txt$/i);
+
+      // filter to valid types only
+      const filteredFiles = targetFiles.filter(
+        f => isExcelFile(f) || isCSVFile(f) || isTextFile(f),
+      );
+
+      this.pendingFiles = this.pendingFiles.concat([...filteredFiles].map(file => ({
+        file,
+        status: 'pending',
+        progress: {},
+        meta: {},
+      })));
+
 
       const promises = this.pendingFiles
         .filter(f => f.status === 'pending')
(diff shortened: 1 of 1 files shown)


sgratzl:
Usually one implements that you are not allowed to drop the file in the first place by rejecting the drop. The file input already filters to show only valid files

waxlamp:
Since I use the file input almost exclusively, I had to change the file filter options in order to test this branch. I admit this is a corner case, but all that means is that someone, somewhere will run into this.

I'm ok with not making the change on this branch, but in that case we should open an issue for it.

sgratzl:
image

subdavis:
I'm usually not a huge fan of toasts and snackbars for error display.  I think it dislocates the user action that caused the error and the error itself.

In this case I think it's OK (and possibly appropriate) because the upload type is already restricted, so users would have to change the rules of their file picker to even get to this error state.  There's also not really a better place to put this error.

web/src/components/Upload.vue
```diff
@@ -69,19 +71,35 @@ export default {
...
-        progress: {},
-        meta: {},
-      })));
-
       const excelMimeTypes = [
         'application/vnd.ms-excel',
         'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
       ];
 
       const isExcelFile = file => excelMimeTypes.includes(file.type) || file.name.match(/\.xlsx?$/i);
```
subdavis: This regex doesn't match the accepted types for the dropzone.  `.xlsx?` will match `.xls` too.

web/src/components/Upload.vue
```diff
@@ -188,7 +210,8 @@ v-layout.upload-component(column, fill-height)
                   :items="dataTypes", label="Type of data",
                   item-text="name", item-value="value")
           v-divider(v-if="idx + 1 < files.length", :key="idx")
-    dropzone.filezone.mx-4.mb-4(:multiple="true", :message="message", @change="onFileChange")
+    dropzone.filezone.mx-4.mb-4(:multiple="true", :message="message", @change="onFileChange",
+        accept=".csv,.xlsx,.txt")
```
subdavis: see above

web/src/components/Upload.vue
```diff
@@ -69,19 +71,35 @@ export default {
...
-      })));
-
       const excelMimeTypes = [
         'application/vnd.ms-excel',
         'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
       ];
 
       const isExcelFile = file => excelMimeTypes.includes(file.type) || file.name.match(/\.xlsx?$/i);
+      const isCSVFile = file => file.type === 'text/csv' || file.name.match(/\.csv$/i);
+      const isTextFile = file => file.type === 'text/plain' || file.name.match(/\.txt$/i);
```
subdavis: Could we move these out of the scope of this method, to the head of the script?

waxlamp:
I agree with @subdavis's comments about toasts and snackbars etc.

PR #269, opened 2019-09-26
Enable TypeScript support

waxlamp:
This is a bit of an experiment in grafting our project onto a TypeScript workflow.

The first commit in this series attempts to swap out our vue-clie setup for one that comes bundled with typescript support.

The remaining commits are my ongoing attempt to do a minimal changeover to typescript. The general procedure is to choose a source file, run `git mv <filename>.js <filename>.ts`, look at the type errors that result on build, then perform minimal updates to the source file to bring it into compliance. I choose a source file by looking at the errors from the build, and picking one that says

```
Could not find a declaration file for module ...
```

and `git mv`ing that file to its `.ts` version.

By default, "minimal updates" means annotating every type that can't be inferred by the engine with `any`; for some cases where it's quite clear what the underlying types are (for instance, in `utils/rangelist.js`), I have been directly supplying those types.

My major goal is to see what it will take to change over to typescript, observe how much of a benefit we will get, and to get buy in from the team. It's my desire to make as few changes as needed that's leading me to do minimal typechecking at this phase.

This approach also gets us a [gradually typed](https://en.wikipedia.org/wiki/Gradual_typing) codebase; the next step would be to set a goal to switch out every `any` for an actual type declaration. This effort will be much longer, but should also be possible with minimal continuing disruption to feature development.

I can certainly use some help in getting the rest of the codebase changed over to gradual typing, and doing so will provide a rough and ready tutorial for typescript besides. So, on this thread, perhaps we can have some sort of protocol for divvying up this work. But before that, I'm hoping for some general feedback on this whole enterprise.

I'll start: typeful programming sometimes requires further up-front work, but pays off dividends in spades; it's a technology that will help us develop a better codebase for Metabolomics, and is something we ought to develop some institutional expertise in as it will also help other projects across the company.


web/src/store/index.ts
```diff
@@ -1,13 +1,22 @@
...
 import {
   convertCsvToRows, RangeList, mapValidationErrors,
-} from '../utils';
-import analyses from '../components/vis/analyses';
-import { plot_types } from '../utils/constants';
-import { CSVService, ExcelService } from '../common/api.service';
+} from '@/utils';
+import analyses from '@/components/vis/analyses';
+import { plot_types } from '@/utils/constants';
+import { CSVService, ExcelService } from '@/common/api.service';
```
subdavis: What do you gain from using `@`?  The only thing I can tell that changes is:

* My editor can no longer produce type hints because it can't resolve these names; and
* My editor can no longer automatically fix all imports if I move a file.

I'm sure I could figure out configuration , but I wonder why this is better.

web/src/store/index.ts
```diff
@@ -99,30 +108,30 @@ const appstate = {
 };
 
 const getters = {
-  dataset: state => id => state.datasets[id],
-  ready: state => id => state.datasets[id] && state.datasets[id].ready,
+  dataset: (state: any) => (id: string) => state.datasets[id],
```
subdavis: I've never used typescript in vuex, but I've always though it looked perilous.  Are you following any particular guide?  Could I read it?

web/src/store/index.ts
```diff
@@ -433,7 +442,7 @@ const actions = {
     }
   },
 
-  async [SET_DATASET_DESCRIPTION]({ commit }, { dataset_id, description }) {
+  async [SET_DATASET_DESCRIPTION]({ commit }: { commit: any }, {dataset_id, description}: { dataset_id: string, description: string }) {
```
subdavis: You've done `{ commit } : any` above and something a little different here.

The people in [this article](https://codeburst.io/vuex-and-typescript-3427ba78cfa8) didn't annotate the first argument to vuex methods at all. What are the benefits of each?

web/src/store/index.ts
```diff
@@ -448,9 +457,11 @@ const actions = {
 
 };
 
-export default new Vuex.Store({
+const store: StoreOptions<RootState> = {
```
subdavis: This is what handles providing type checking to interaction with `store` from within mutations, getters, etc right?

Do you know what happens when you set stuff with `Vue.set(store, 'key', 'val')` ?  I would imagine type validations are just not caught.

I imagine the main purpose of types in vuex is for validating external interactions, such as accessing the store's proerties from `mapState` -- is that right?

web/src/store/index.ts
waxlamp: I think it's better for two reasons:
1. It's faster to type because you don't have to think about which level of the directory hierarchy you're in.
2. It enables moving files to other places more readily, in case of a code reorg.

You would still use relative paths for files that are meant to be bundled together in the same directory, etc. (i.e., ones that would move along with the source file in case of a reorg).

Your editor not supporting this is a big deal. I am surprised though: I didn't invent this `@' thing; it comes with Vue CLI projects out of the box.

If this causes major problems for the dev team we shouldn't use it (although I do really like it).

web/src/store/index.ts
waxlamp: Do you mind clarifying why using typescript with vuex would be dangerous?

I'm only using a styleguide implicitly, via [TSLint](https://github.com/palantir/tslint), which was set up by Vue CLI's config for typescript.

(In this particular case, the forced parenthesization of a single argument in an arrow function is actually justified: typescript accepts `x => ...`, `(x) => ...`, and `(x: T) => ...`, but not `x: T => ...`.)

web/src/store/index.ts
subdavis: I'll see if I can get my editor to like it.  Shouldn't be too hard.

web/src/store/index.ts
waxlamp: The `any`s in both instances are placeholders and therefore equally "bad" (i.e., must be replaced eventually by an actual type). `{ commit }: any` is "more equal" because it's less precise. But in this case, both type declarations have the exact same semantics.

So the benefit of the latter approach is that it already has the correct shape for the eventually correct typing for that expression. Better even than that is probably to do something like

```typescript
interface CommitArg {
  commit: ...;
}
```

and then type the expression as `{ commit }: CommitArg`.

I assume that article didn't type the first argument either because they were focused on demonstrating something else, or the TS compiler was able to determine the type of that argument solely from usage in their example codebase. I stuck in a placeholder typing because the compiler told me it couldn't infer the type. (Though, honestly, I believe it's better to type as much as possible in order to avoid problems with type-compatible but semantically forbidden usages later in the lifecycle of the codebase.)

web/src/store/index.ts
waxlamp: > This is what handles providing type checking to interaction with `store` from within mutations, getters, etc right?
> 
> Do you know what happens when you set stuff with `Vue.set(store, 'key', 'val')` ? I would imagine type validations are just not caught.
> 

I admit, I don't understand some of the deeper magic in Vue's specific type bindings (there's some magic, for instance, that allows TS to recognize props and computed properties for the value returned by `Vue.extend()`).

An experiment with a poorly typed `Vue.set()` invocation seems to be in order.

> I imagine the main purpose of types in vuex is for validating external interactions, such as accessing the store's proerties from `mapState` -- is that right?

Yes this sounds right (though it is possible that the "internal" interactions might also be checked).

web/src/router/index.ts
```diff
@@ -62,7 +62,7 @@ export const routes = [
         component: AnalyzeData,
         props: true,
       },
-      ...analyses.map(({ path, shortName: name, component }) => ({
+      ...analyses.map(({ path, shortName: name, component }: { path: string, shortName: string, component: any }) => ({
```
sgratzl: isn't it: `component: Vue`?

web/tslint.json
```diff
@@ -0,0 +1,19 @@
+{
```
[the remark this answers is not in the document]
waxlamp: Good to know. It looks like they're recommending ESLint instead but I'll need to do more research to find whether and how to switch over.

PR #277, opened 2019-09-30
generate full validated table

### LINKED ISSUE #273: Enable preprocessed table download as CSV
Enable a control for downloading the table used for the PCA analysis. One potential place for a download button is at the bottom of the normalization controls panel.

Ideally the downloaded table would have all input data/metadata/group columns in the original data. I'm unsure if this is available directly or if it would need to be pulled together based on separate info. The server would probably be the place to produce the full CSV file using official pandas serialization.

sgratzl:
closes #273

I can regenerate the full table almost from the derived subsets (groups, metadata, ...)


viime/views.py
```diff
@@ -499,6 +499,14 @@ def set_scaling_method(validated_table):
         raise
 
 
+@csv_bp.route('/csv/<uuid:csv_id>/validate/download', methods=['GET'])
+@load_validated_csv_file
+def download_validated_csv_file(validated_table):
+    fp = BytesIO(validated_table.table.to_csv(header=False, index=False).encode())
+    return send_file(fp, mimetype='text/csv', as_attachment=True,
+                     attachment_filename=validated_table.name)
```
jbeezley: Should we append `.csv` in the attachment name?

jbeezley:
For what this returns, I think the implementation is fine.  I'm still unsure if this is what we really want to send back.

I would propose an alternative that would just tile the four main tables to provide a more normalized version of the table.

e.g. columns created by concatenating in order:
* primary key
* groups
* sample metadata
* processed metabolite data

rows would be:
* header
* metabolite metadata
* processed metabolite data

The masked rows/columns would not appear.  I think the implementation would be pretty easy (np.block, for example).  We might want to ask users what they would prefer, but we would need a simplified way of explaining it.

sgratzl:
> I would propose an alternative that would just tile the four main tables to provide a more normalized version of the table.
> 
> e.g. columns created by concatenating in order:
> 
>     * primary key
> 
>     * groups
> 
>     * sample metadata
> 
>     * processed metabolite data
> 
> 
> rows would be:
> 
>     * header
> 
>     * metabolite metadata
> 
>     * processed metabolite data
> 
> 
> The masked rows/columns would not appear. I think the implementation would be pretty easy (np.block, for example). We might want to ask users what they would prefer, but we would need a simplified way of explaining it.

@jeffbaumes your call

jeffbaumes:
I'm not sure I understand all the differences between the current implementation and @jbeezley 's suggestion, but I believe @jbeezley 's idea would work well. That is, don't worry about preserving original column order (though I expect in most cases the data would originally be ordered close to that anyway), and leave out masked rows/columns entirely.

jbeezley:
The primary difference is that the current implementation keeps the original row/column order.

jeffbaumes:
Ok, then yes as long as the code gets less complex and not more complex, I'm in favor of @jbeezley 's suggestion.

viime/models.py
```diff
@@ -712,6 +712,30 @@ def measurements(self):
...
 
+    @property
+    def table(self):
+        """
+        mimics the .table from the CSVFile but with validated data
+        """
+        return _generate_validated_table(self)
+
+
+def _generate_validated_table(valdidated_table):
```
jbeezley: valdidated -> validated

web/src/common/api.service.js
```diff
@@ -69,6 +69,14 @@ export const CSVService = {
     return ApiService.download('csv', slug);
   },
 
+  downloadUrl(slug) {
```
subdavis: This function is never called.

sgratzl: `download(slug) {` neither but it still there

subdavis: Then can we remove the dead code?

PR #278, opened 2019-09-30
export svg to pngs

### LINKED ISSUE #270: Add logic to download SVG plots as PNG
I don't know the mechanism to do this in general, or if it is even possible/feasible, so the start of this task would be feasibility analysis. Ideally this could be done in-browser with any SVG-based visualization (a utility function that takes an SVG element and initiates a PNG download).

sgratzl:
closes #270 

e.g:

PCA Score Plot

Notes
 * the generated SVG file should contain all CSS rules that it is affected by
 * However, the *external* font isn't used due to security restrictions by the browser. It is part of the SVG but included externally (there is no other way)

### DIFF
--- a/web/src/components/vis/VisTile.vue
+++ b/web/src/components/vis/VisTile.vue
@@ -1,5 +1,6 @@
 <script>
 import { format } from 'd3-format';
+import { downloadSVG } from '../../utils/exporter';
 
 export default {
   props: {
@@ -8,6 +9,10 @@ export default {
       default: '',
     },
     loading: {
+      default: false,
+      type: Boolean,
+    },
+    svgDownload: {
       default: false,
       type: Boolean,
     },
@@ -43,6 +48,12 @@ export default {
     setScaleIndex(value) {
       this.scaleIndex = Math.max(Math.min(Number.parseInt(value, 10), this.scales.length - 1), 0);
     },
+    downloadSVG() {
+      const svg = this.$el.querySelector('svg');
+      if (svg) {
+        downloadSVG(svg, this.title);
+      }
+    },
   },
 };
 </script>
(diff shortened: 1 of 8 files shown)


web/src/utils/exporter.js
```diff
@@ -0,0 +1,74 @@
...
+
+  // based on http://bl.ocks.org/biovisualize/8187844
+  const copy = svgElement.cloneNode(true);
+  // proper bg
+  copy.style.backgroundColor = 'white';
+  // inject font
+  copy.style.fontFamily = 'Barlow Condensed, sans-serif';
+  copy.insertAdjacentHTML('afterbegin', `<style>
+  @import url("https://fonts.googleapis.com/css?family=Barlow+Condensed");
+  </style>`);
```
jeffbaumes: Does attempting to import the font have any effect? Seems it will just revert to sans-serif.

jeffbaumes:
This seems to work well for me. I wish the font would work so it was WYSIWYG. I'll add an issue to choose a chart font that is built-in and consistent across OSes and also works well with the app, since I think WYSIWYG is important.

@subdavis can give his thoughts but the code looks reasonable to me.

web/src/utils/exporter.js
sgratzl: when you switch to downloading the SVG image, you can see the font

sgratzl:
I solve the problem by inlining the font as a data uri

PCA Score Plot (2)

web/src/components/vis/VisTile.vue
```diff
@@ -1,5 +1,6 @@
 <script>
 import { format } from 'd3-format';
+import { downloadSVG } from '../../utils/exporter';
```
waxlamp: I think we should prefer `@/utils/explorer' for file imports that don't have a need to be relative to the current file.

web/src/utils/exporter.js
```diff
@@ -0,0 +1,87 @@
+// eslint-disable-next-line import/no-webpack-loader-syntax
+import font from '!url-loader?limit=undefined!@openfonts/barlow-condensed_all/files/barlow-condensed-all-400.woff2';
```
waxlamp: Can we put a loader rule into `vue.config.js` instead of doing it inline?

sgratzl: I need to force the loader to inline. By default it would fallback to the file-loader since it would consider the font too large to inline

web/src/components/vis/VisTile.vue
sgratzl: see also https://github.com/girder/viime/pull/269#discussion_r330278495

jeffbaumes: Seems that after we resolve what we want to do here, someone can go through and make sure @ is consistently used (or consistently not used). Don't think this blocks this PR.

PR #279, opened 2019-09-30
Hash based rangelist

subdavis:
@sgratzl this pr includes a few other suggested fixes.  PTAL.  If you disagree, I can revert anything you think is necessary.  Once this is merged, I'll approve #264

### DIFF
--- a/web/src/utils/rangelist.js
+++ b/web/src/utils/rangelist.js
@@ -1,29 +1,18 @@
+import Vue from 'vue';
 /**
  * An inefficient rangelist
  * because there shouldn't be a large fragmented list of ranges.
  */
 export default class RangeList {
   constructor(initial = []) {
-    this.members = initial;
+    this.members_ = {};
+    initial.forEach(i => this.add(i));
   }
 
-  /**
-   * Binary search for el
-   * @param {Number} el element to find
-   */
-  _find(el) {
-    let lowIndex = 0;
-    let highIndex = this.members.length - 1;
-    while (lowIndex <= highIndex) {
-      const midIndex = Math.floor((lowIndex + highIndex) / 2);
-      if (this.members[midIndex] === el) {
-        return midIndex;
-      } if (this.members[midIndex] < el) {
-        lowIndex = midIndex + 1;
-      } else {
-        highIndex = midIndex - 1;
-      }
-    } return null;
+  get members() {
+    return Object.keys(this.members_)
+      .map(m => parseInt(m, 10))
+      .sort((a, b) => a - b);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


sgratzl:
why don't use use a regular https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set? Morever, with a set based approach you loose the order of selection in contrast to a list

subdavis:
Yep, that would make way more sense.

subdavis:
@sgratzl turns out there is a reason not to use a Set. I found https://github.com/vuejs/vue/issues/2410#issuecomment-318487855

> Sets and Maps are not reactive in vue

I think this implementation is simple enough for now.  We'd have to wrap a set with some other class to get the `first` and `last` properties.

web/tests/rangelist.spec.js
```diff
@@ -1,26 +1,6 @@
 import RangeList from '../src/utils/rangelist';
 
 describe('RangeList', () => {
-  it('Produces non-overlapping ranges', () => {
```
sgratzl: why did you delete all those tests?

subdavis: Because of the implementation change, this test case no longer applies.  Overlap is not meaningful for a hash-based approach:

Before, `this.members = [ 1, 2, 2, 3, 4, 5 ]` was a possible edge case this test guarded against.

web/src/store/index.js
```diff
@@ -227,7 +227,7 @@ const mutations = {
     } else if (event.ctrlKey && axis === type) {
       ranges.add(idx);
     } else if (!event.ctrlKey && !event.shiftKey) {
-      state.datasets[key].selected.ranges = new RangeList([idx]);
+      state.datasets[key].selected.ranges = new RangeList({ [idx]: true });
```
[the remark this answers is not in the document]
subdavis: Agreed.

web/src/utils/rangelist.js
```diff
@@ -1,29 +1,15 @@
...
-      if (this.members[midIndex] === el) {
-        return midIndex;
-      } if (this.members[midIndex] < el) {
-        lowIndex = midIndex + 1;
-      } else {
-        highIndex = midIndex - 1;
-      }
-    } return null;
+  get members() {
+    return Object.keys(this.members_).map(m => parseInt(m, 10));
```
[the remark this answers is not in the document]
subdavis: The usages don't need sorting, but I still think that's a reasonable expectation.

web/src/utils/rangelist.js
```diff
@@ -1,29 +1,18 @@
...
-      } if (this.members[midIndex] < el) {
-        lowIndex = midIndex + 1;
-      } else {
-        highIndex = midIndex - 1;
-      }
-    } return null;
+  get members() {
+    return Object.keys(this.members_)
+      .map(m => parseInt(m, 10))
+      .sort();
```
sgratzl: this will sort alphabetically not numerically

web/tests/rangelist.spec.js
sgratzl: imo this is still a valid test. I don't get why it shouldn't pass anymore. And the case you are describing is still part of the specification of the RangeList class.

PR #291, opened 2019-10-02
denser table

### LINKED ISSUE #289: Fit more p-value data on screen for Wilcoxon and ANOVA
Condense p-value table rows. The default number of visible rows could also be changed to the full amount to avoid unused empty screen space.

We could try using the data table view, but that does not enable sorting rows by column value so I think we should stick to the vuetify table for now.

sgratzl:
closes #289

image

### DIFF
--- a/web/src/components/vis/MetaboliteTable.vue
+++ b/web/src/components/vis/MetaboliteTable.vue
@@ -22,6 +22,9 @@ export default {
   data() {
     return {
       format: format('.4e'),
+      pagination: {
+        rowsPerPage: -1,
+      },
     };
   },
 
@@ -34,17 +37,29 @@ export default {
 </script>
 
 <template lang="pug">
-v-data-table.elevation-1(:headers="headers", :items="items", disable-initial-sort,
-    item-key="Metabolite")
+v-data-table.elevation-1.main(:headers="headers", :items="items", disable-initial-sort,
+    item-key="Metabolite", :pagination.sync="pagination")
   template(v-slot:items="props")
-    td {{ props.item.Metabolite }}
-    td.text-xs-right(v-for="c in headers.slice(1)",
+    td.cell {{ props.item.Metabolite }}
+    td.cell.text-xs-right(v-for="c in headers.slice(1)",
         :class="isInteresting(props.item[c.value]) ? 'highlight' : ''")
       | {{ format(props.item[c.value]) }}
 </template>
 
-<style lang="scss" scoped>
+<style scoped>
 .highlight {
   background-color: #ffadad;
+}
+
+.main {
+  height: 0;
+}
+
+.main >>> tr {
+  height: 25px;
+}
+.main >>> td.cell {
+  height: 25px;
+  padding: 2px 7px;
 }
 </style>


web/src/components/vis/MetaboliteTable.vue
```diff
@@ -22,6 +22,9 @@ export default {
   data() {
     return {
       format: format('.4e'),
+      pagination: {
+        rowsPerPage: -1,
+      },
```
waxlamp: Does this just mean "as many pages as possible"?

web/src/components/vis/MetaboliteTable.vue
```diff
@@ -34,17 +37,29 @@ export default {
...
-    td {{ props.item.Metabolite }}
-    td.text-xs-right(v-for="c in headers.slice(1)",
+    td.cell {{ props.item.Metabolite }}
+    td.cell.text-xs-right(v-for="c in headers.slice(1)",
         :class="isInteresting(props.item[c.value]) ? 'highlight' : ''")
       | {{ format(props.item[c.value]) }}
 </template>
 
-<style lang="scss" scoped>
+<style scoped>
```
waxlamp: Why get rid of SCSS here?

web/src/components/vis/MetaboliteTable.vue
```diff
@@ -34,17 +37,29 @@ export default {
...
 .highlight {
   background-color: #ffadad;
 }
+
+.main {
+  height: 0;
+}
+
+.main >>> tr {
```
waxlamp: What does this selector do?

sgratzl: https://vue-loader.vuejs.org/guide/scoped-css.html#deep-selectors

web/src/components/vis/MetaboliteTable.vue
sgratzl: https://v15.vuetifyjs.com/en/components/data-tables

```js
{
  descending: boolean",
  page: number",
  rowsPerPage: number // -1 for All",
  sortBy: string",
  totalItems: number"
}
```

waxlamp: A comment in our code might be a good idea.

web/src/components/vis/MetaboliteTable.vue
sgratzl: SCSS doesn't support the deep selector

web/src/components/vis/MetaboliteTable.vue
```diff
@@ -34,17 +37,29 @@ export default {
 </script>
 
 <template lang="pug">
-v-data-table.elevation-1(:headers="headers", :items="items", disable-initial-sort,
-    item-key="Metabolite")
+v-data-table.elevation-1.main(:headers="headers", :items="items", disable-initial-sort,
```
subdavis: Not really worth changing here, but I tend toward more descriptive class names.  `metabolite-table` for instance.

web/src/components/vis/MetaboliteTable.vue
```diff
@@ -34,17 +37,29 @@ export default {
...
 }
+
+.main {
+  height: 0;
+}
+
+.main >>> tr {
+  height: 25px;
+}
+.main >>> td.cell {
```
subdavis: I'm a big fan of this.  I usually end up using vuetify's internal class names and it bites me when they change.  I like that this is an alternative option.

PR #293, opened 2019-10-02
add optional help icon with modal popup

### LINKED ISSUE #288: Add help icon to plots
Clicking the icon would open a help dialog describing the plot and plot options.

Would this be another named slot in VisTile? I could see the documentation for some plots or analyses being quite lengthy, could include embedded images, etc. so we probably want a slot that accepts a full component, as opposed to assuming the help will be a text-only string.

Placeholder help would be fine with this PR, then we can fill in the help later.

sgratzl:
closes #288

image


web/src/utils/vuetifyConfig.js
```diff
@@ -42,6 +42,7 @@ export default {
     eyedropper: 'mdi-eyedropper',
     group: 'mdi-group',
     header: 'mdi-key-variant',
+    help: 'mdi-help',
```
jeffbaumes: I like help-circle a bit better. The current ? is too big for my taste.

web/src/components/vis/VisTile.vue
```diff
@@ -52,6 +56,18 @@ v-flex(shrink=1, :class="scaleClass")
   .white.rounded.relative
     v-toolbar.primary.darken-3.top-rounded(dark, flat, dense)
       v-toolbar-title {{ title }}
+      v-toolbar-items(v-if="hasHelp()")
+        v-dialog(v-model="showHelp", max-width="33vw")
+          template(v-slot:activator="{ on }")
+            v-btn(v-on="on", icon)
+              v-icon {{ $vuetify.icons.help }}
+          v-card
+            v-card-title {{ title }}
```
jeffbaumes: I suggest updating the title to this:

```
            v-card-title
              h3.headline {{ title }}
```

This extra styling is not needed when using Vuetify 2, but I guess we are still on 1.x.

web/src/components/vis/VisTile.vue
```diff
@@ -48,6 +49,9 @@ export default {
     setScaleIndex(value) {
       this.scaleIndex = Math.max(Math.min(Number.parseInt(value, 10), this.scales.length - 1), 0);
     },
+    hasHelp() {
+      return this.$slots.help != null;
```
subdavis: `this.$slots.help` cannot be null, so this check relies on coercion without appearing to do so.  Either `this.$slots.help !== undefined` (no coercion) or `!!this.$slots.help` (obvious coercion) would probably be better.

sgratzl: `this.$slots.help != null` checks for both `!== null` and `!== undefined` at the same time. That's why I just used `!=` and not `!==`

sgratzl: note: and `!= null` is the only time a good linter will allow not to use `!==` for comparison

PR #295, opened 2019-10-02
Sgratzl/columnselection

### LINKED ISSUE #283: Global selection read/write endpoint
Create a way to store/retrieve a column selection per table.

### LINKED ISSUE #284: Global selection controls on ANOVA and Wilcoxon panels

### LINKED ISSUE #292: Condense table in horizontal direction
The denser table in #291 looks great, but I wonder if we can eliminate the horizontal dead space as well. It would make the tables much easier to read and manipulate (see the screencap in #291).

sgratzl:
based on #291 

closes #283, closes #284, closes #292 

image


viime/models.py
```diff
@@ -297,6 +298,8 @@ class CSVFileSchema(BaseSchema):
     columns = fields.List(fields.Nested('TableColumnSchema', exclude=['csv_file']))
     rows = fields.List(fields.Nested('TableRowSchema', exclude=['csv_file']))
 
+    selected_columns = fields.List(fields.Str(), dump_only=True, missing=[])
```
jbeezley: `missing=list` to avoid using a mutable object as a default.

viime/models.py
```diff
@@ -74,6 +74,7 @@ class CSVFile(db.Model):
     imputation_mnar = db.Column(db.String, nullable=False)
     imputation_mcar = db.Column(db.String, nullable=False)
     meta = db.Column(JSONType, nullable=False)
+    selected_columns = db.Column(db.PickleType, nullable=True)
```
jbeezley: Cool, I didn't know sqlalchemy support pickling out of the box.

web/src/components/vis/WilcoxonPlotTile.vue
```diff
@@ -3,6 +3,7 @@ import WilcoxonPlot from '@/components/vis/WilcoxonPlot.vue';
 import VisTileLarge from '@/components/vis/VisTileLarge.vue';
 import ToolbarOption from '../ToolbarOption.vue';
 import plotData from './mixins/plotData';
+import { SET_DATASET_SELECTED_COLUMNS } from '../../store/actions.type';
```
subdavis: Perhaps change this variable name to `SET_ANALYSIS_SELECTED_COLUMNS`.  I got this confused with `SET_SELECTION` for the cleanup table.

web/src/store/index.js
```diff
@@ -162,6 +164,7 @@ const mutations = {
         name,
         description,
         created: new Date(created),
+        selectedColumns: selectedColumns || [],
```
subdavis:
```suggestion
        selectedColumns: dataset.selected_columns || [],
```

web/src/store/index.js
```diff
@@ -151,7 +153,7 @@ const mutations = {
    */
   [SET_DATASET_DATA](state, { data }) {
     const {
-      id, name, size, created, description,
+      id, name, size, created, description, selected_columns: selectedColumns,
```
subdavis:
```suggestion
      id, name, size, created, description,
```

subdavis: I don't think there's a need to extract this at the top unless the payload key matches the dataset property key exactly.

jeffbaumes:
@subdavis There was a bit of post-meeting discussion between me, @sgratzl , and @jbeezley. The set of metabolites of interest is the major output of this application (e.g. to download to downstream analyses like pathway analysis). I think it's true that the selection will be set/used primarily within the analysis pages, but we could for example have those highlighted/changed in the PCA views in the transform table page. So in a sense the selection is truly global per dataset. So I think @sgratzl's naming of `SET_DATASET_SELECTED_COLUMNS ` is a decent choice. Perhaps the existing `SET_SELECTION` is the name that should include additional qualifiers that it is associated with cleanup.

subdavis:
> Perhaps the existing SET_SELECTION is the name that should change to include additional qualifiers that it is associated with cleanup.

Given that context, renaming `SET_SELECTION` commit and `selected` property to include `cleanup` in the name sounds totally reasonable.

subdavis:
@sgratzl `dataset` should have been `data`.  Sorry :( That's actually kind of a decent lesson about the dangers of suggesting inline edits.

PR #299, opened 2019-10-02
Column Selection in Correlation Network

### LINKED ISSUE #285: Use global selection in correlation network view

sgratzl:
closes #285 

image

### DIFF
--- a/viime/analyses.py
+++ b/viime/analyses.py
@@ -1,5 +1,5 @@
 from itertools import combinations
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, Optional
 
 import pandas as pd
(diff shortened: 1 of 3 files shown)


subdavis:
Ooh, this is cool.

jeffbaumes:
Yeah, if I have the correct intuition on the central role of an "interesting metabolite" set, this type of cross-analysis selection stuff is going to be key to selling this thing to the community.

Thanks for the spot-on suggestion @sgratzl.

jeffbaumes:
@subdavis Interesting link. I don't think we should go down the optimization route on this view just yet, but we could track this with an issue.

PR #276, opened 2019-09-30
add pan and zoom to correlation plot

sgratzl:
closes #271

image

PR #300, opened 2019-10-02
Sgratzl/correlationdrag

### LINKED ISSUE #286: Allow users to pull nodes around

### LINKED ISSUE #305: Fix padding/margin on correlation graph
There is unwanted padding/margin at the top:

image

sgratzl:
closes #286 and closes #276 and closes #305

image


sgratzl:
> I like the pinning. Is there a way to unlock pinned nodes? I could envision clicking on a pinned node would unlock it, if that can be distinguished from a drag event.

done

also the zooming

image

image

subdavis:
* Perhaps we need to add an explanation of the available controls on this page somewhere.
* Dragging stops working when you change the method.

sgratzl:
>     * Perhaps we need to add an explanation of the available controls on this page somewhere.  Just something to let the user know they _can_ pan, zoom, drag, and something about how pinning works.

see also #317, we need a general concept how to deal with such things

>     * Dragging breaks when you change the method (usually)

fixed

PR #304, opened 2019-10-02
Imputation Tab

### LINKED ISSUE #250: Generalize DataTable.vue to also render the validated table
So, far the datatable is optimized in rendering the raw dataset. However, it could be extended to also render the validated data (or a subset of it, like just the measurements) as needed by #238

sgratzl:
closes #238, closes #250 

image


web/src/components/Pretreatment.vue
```diff
@@ -90,6 +90,13 @@ v-layout.pretreatment-component(row, fill-height)
                     v-icon.pl-1(small, @click="", v-on="on") {{ $vuetify.icons.info }}
                   span {{ problemData.description }}
 
+            v-list-tile.ml-4.my-1.px-0.small-tile(
+                :class="{ active: $router.currentRoute.name === 'Impute Table' }",
+                @click="$router.push({ path: `/pretreatment/${dataset.id}/cleanup/impute` })")
+              v-list-tile-title.small-tile.pl-2
+                v-icon.pr-1.middle {{ $vuetify.icons.tableEdit }}
```
jeffbaumes: Perhaps make an `impute` icon which is `mdi-file-table-box-outline`?

sgratzl: we don't have this specific icon in our iconfont

jeffbaumes: Hm, ok @jtomeck may need to rescue us here. I'm ok with keeping this icon for now.

subdavis: @sgratzl Are you saying it isn't available in version 3.5.95, or that it needs to be added in `utils/vuetifyConfig.js`?  I see that icon in @mdi/font so we should be able to use it.

sgratzl: there is no CSS class defined in `@mdi/font/materialdesignicons.css` with the name `mdi-file-table-box-outline` in my version 3.5.95

jeffbaumes: Looks like it's in master for mdi, maybe we just `yarn update @mdi/font` and it will be there?

https://github.com/Templarian/MaterialDesign-Webfont/blob/master/css/materialdesignicons.css

jbeezley:
I think you have rows/columns mixed up somewhere because the wrong cells are getting highlighted.

sgratzl:
> I think you have rows/columns mixed up somewhere because the wrong cells are getting highlighted.

good spot, fixed

waxlamp:
Do we have a dataset with missing values that we could use to test branches such as this?

jeffbaumes:
Per @subdavis suggestion, I see that you extracted the requested code into a separate file for maintainability. I think this is clear to merge.

PR #308, opened 2019-10-03
Sgratzl/heatmapfixes

### LINKED ISSUE #303: Misc Heatmap bugs.
First of all, sorry for approving #266  when I hadn't run it locally.

* The mouseover animation is a bit too heavy and causes the UI to lock up when you mouse around on the view with the debug console open.  I know this isn't a normal operating condition, but if something is slow enough to break when debug is open, it probably won't run well on cheap hardware.  Should probably dig deeper into the heft of the render loop and test with much larger file. A set of test files as mentioned in #281 would also help here.  Might also have to sacrifice some of the nice highlighting for speed.
* It looks like you're redrawing the highlighting after every mouse move.  Could we debounce it by not firing the redraw if the highlighted cell hasn't changed from the last mousemove event?
* the resize animation is quite slow.  If you drag the page around, it takes the tree a while to catch up.  Is it possible to drop the animation and just render it immediately?
* the animation for adding and removing rows and columns is also pretty slow: the heatmap redraws immediately and then the tree struggles to catch up.  
* resizing the page causes the heatmap to disappear.  It seems to come back on mouseover.
* there's some odd behavior with the selection cursor when width < height.  Screenshot below.
* I think the plus and minus buttons may be backward.  `+` hides things and `-` adds them back.

Screenshot from 2019-10-02 16-13-32

Also included a flame graph for mouseover redrawing.

Screenshot from 2019-10-02 16-03-26

Would it make sense to revert #266 until these issues are fixex?  Again, sorry for not trying this out before I approved it.  

cc @sgratzl

sgratzl:
closes #303 

notes
 * no more animation cause there is no performant way to fix it and the canvas one is too difficult to do


subdavis:
> It looks like you're redrawing the highlighting after every mouse move. Could we debounce it by not firing the redraw if the highlighted cell hasn't changed from the last mousemove event?

Looks like this is still an issue.  You could pretty dramatically reduce the redraw cost of highlighting with #329 or something like that.

Looks like all the other bugs were addressed.

subdavis:
I agree it isn't ideal, I agree.  The coloring suggestion sounds great -- could you open an issue for that?

IMO, for this particular case, bad animation is worse than no animation.

waxlamp:
> I agree it isn't ideal, I agree. The coloring suggestion sounds great -- could you open an issue for that?

Check out #338.

PR #315, opened 2019-10-04
Sgratzl/heatmapfocus

### LINKED ISSUE #312: "Zoom to cluster" in heatmap view
The heatmap currently enables collapsing a clusters. Zoom to cluster, where only that cluster is visible, would also be useful. There would need to be a way to zoom back out to the full data.

Unsure how to work the UI such that both collapse and zoom are possible, would we need an interaction mode toggle?

sgratzl:
based on #308 

closes #312 

so far when the user clicks while holding the ctrl-button, it will focus on the node instead of collapsing it. @jeffbaumes what do you think?

image

### DIFF
--- a/web/src/components/vis/Heatmap.vue
+++ b/web/src/components/vis/Heatmap.vue
@@ -220,7 +220,7 @@ export default {
         r.on('click', (d) => {
           if (wrapper.focus === d.data) {
             wrapper.focus = null;
-          } else if (event.ctrlKey) {
+          } else if (event.ctrlKey || event.shiftKey) {
             wrapper.focus = d.data;
           } else if (wrapper.collapsed.has(d.data)) {
             wrapper.collapsed.delete(d.data);
--- a/web/src/components/vis/Heatmap.vue
+++ b/web/src/components/vis/Heatmap.vue
@@ -245,6 +245,7 @@ export default {
       });
       inner.select('title').text(d => d.data.name);
       inner.classed('collapsed', d => collapsed.has(d.data));
+      inner.classed('focussed', d => wrapper.focus === d.data);
 
       inner.attr('transform', d => `translate(${d.x},${d.y})`);
     },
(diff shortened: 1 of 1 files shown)


jeffbaumes:
On my mac, Ctrl-click brings up a the browser context menu, and the event does not seem to get to the page to trigger the zoom.

sgratzl:
re mopdifier

> On my mac, Ctrl-click brings up a the browser context menu, and the event does not seem to get to the page to trigger the zoom.

re  modifiers, mac behaves differently. 

 * https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/ctrlKey
 * https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/altKey
 * https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/metaKey
 * https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/shiftKey

for now I also listen for `shift+click` events

web/src/components/vis/Heatmap.vue
```diff
@@ -220,9 +237,15 @@ export default {
...
-      inner.select('text').html(d => (collapsed.has(d.data) ? MDI_PLUS_CIRCLE : MDI_MINUS_CIRCLE));
+      inner.select('text').html((d) => {
+        if (wrapper.focus === d.data) {
+          return MDI_STAR_CIRCLE;
+        }
+        return collapsed.has(d.data) ? MDI_PLUS_CIRCLE : MDI_MINUS_CIRCLE;
+      });
       inner.select('title').text(d => d.data.name);
       inner.classed('collapsed', d => collapsed.has(d.data));
+      inner.classed('focussed', d => wrapper.focus === d.data);
```
jeffbaumes: `focussed` -> `focused`

Both are acceptable, but the latter is much more common in American English.

https://english.stackexchange.com/questions/4791/focussed-or-focused-rules-for-doubling-the-last-consonant-when-adding-ed

web/src/components/vis/Heatmap.vue
```diff
@@ -445,7 +468,8 @@ export default {
   fill: orange;
 }
 
-.nodes >>> .collapsed {
+.nodes >>> .collapsed,
+.nodes >>> .focussed {
```
jeffbaumes: Same.

PR #316, opened 2019-10-04
Sgratzl/heatmapfix2

### LINKED ISSUE #313: More heatmap improvements
* [x] The lines in the dendrogram could be half as thick, they look chunky to me.
* [x] The font would look better if always consistent between rows and columns. Perhaps using the min of the current row/col font size would make sense to continue to avoid overlap.
* [x] Since they take up so much room and may not always be useful, enable toggling the dendrogram (toggling row and column dendrograms separately would be good). This would probably require adding side panel for the view. We would eventually want to have other controls, like filtering by global metabolite selection, so it may be fine to ad the side panel now.
* [x] For a clean view (e.g. for downloading for publication) it may be good to be able to set an aspect ratio and/or size of the heatmap boxes. This would require a scrollable view amd some other considerations so I'm leaving this up to @sgratzl to decide a good strategy here (if there is one).

sgratzl:
base on #308 

closes #313 

I added three different layout modes: 
 * auto ... as before
 * square cells ... ensure that cells are squared
 * square matrix ... ensure that the matrix is a square

image

image


jeffbaumes:
I really like the options you give for cell shape. What's a good solution for the bigger table use case? Performance is fine, but I cannot read the text on the sides. The title text is helpful but I'm unsure how the user can quickly scan the rows/columns.

image

We could probably leave this till a project meeting where we can discuss with Tom/Dan on what they may want to see. Some other high-level context, like using different color scales for different sample groups or the selected metabolite set, may provide the needed context when looking at a large table, since I can't think of how to present all the labels at once.

jeffbaumes:
Maybe coloring the horizontal leaf-level dendrogram lines and/or sample names by group could help, so that the interior heatmap color scale can remain consistent.

sgratzl:
we have to think about how to integrate the new column selection into it. As far as I understood the most common case will be to show only the selected subset.

PR #319, opened 2019-10-07
Group Level Metainformation

### LINKED ISSUE #318: Global group colors to use across analyses
This would be like the recently-added metabolite set, where a dataset would have store a configurable set of colors for the groups in the dataset. Unlike metabolite selection, which would primarily be set through analysis views, I see group colors being set in a central place, perhaps in the dataset details page for now. All views would use this scale when coloring samples, right now just the PCA scores plot.

I could see the colors being set automatically from a fixed categorical scale when the dataset groups are first defined (or change by changing the group column), and then allow customization by the user.

Should we include an optional shape as well for color deficient users?

sgratzl:
closes #318 

creates the data model and an editor in the UI for editing group levels

image

related
 * use the actual color mapping in various plots: #321 
 * use the colors in the CleanUp table: #322
 * use the colors in the Heatmap: #330 
 * ?


viime/views.py
```diff
@@ -279,6 +280,27 @@ def set_csv_file_description(csv_id, description):
...
 
 
+@csv_bp.route('/csv/<uuid:csv_id>/group-levels', methods=['PUT'])
+@use_kwargs({
+    'group_levels': fields.List(webargs.fields.Nested({
+        'name': fields.Str(required=True),
+        'label': fields.Str(required=True),
+        'description': fields.Str(missing=None),
+        'color': fields.Str(required=True),
+    }), required=True)
```
jbeezley: webargs works with marshmallow, so I think you can replace this with:
`fields.List(GroupLevelSchema(), required=True)`

sgratzl: the `GroupLevelSchema` requires a `csv_file_id` which is not given by the request parameters

jbeezley: How about `GroupLevelSchema(exclude=['csv_file_id'])`?

viime/models.py
```diff
@@ -66,6 +67,25 @@ def make_object(self, data, **kwargs):
         return data
 
 
+class GroupLevel(db.Model):
```
subdavis: A GroupLevel sounds like a quantitative piece of information, like a threshold.  What does GroupLevel mean?

sgratzl: coming from the R world: if you want to get the different unique values of a factor you call `levels()`

see also https://www.rdocumentation.org/packages/base/versions/3.6.1/topics/levels and https://www.rdocumentation.org/packages/base/versions/3.6.1/topics/factor

subdavis: Ah neat thanks!

PR #321, opened 2019-10-07
apply group colors to score plot

sgratzl:
based on #319 

changes the score plot to use the defined group colors

### DIFF
--- a/web/src/components/vis/ScorePlot.vue
+++ b/web/src/components/vis/ScorePlot.vue
@@ -13,7 +13,7 @@
   .tooltip(ref="tooltip")
 </template>
 
-<style scoped lang="scss">
+<style scoped>
 div.tooltip {
   position: fixed;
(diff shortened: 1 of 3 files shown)


jeffbaumes:
When I tried just now, score plot ellipses did not react to color changes (dots did).

web/src/components/vis/ScorePlot.vue
```diff
@@ -342,14 +330,14 @@ export default {
           .attr('cy', 0)
           .attr('r', 1)
           .attr('style', 'fill: none; stroke: black;')
```
jeffbaumes: This `stroke:black` on the `<circle>` is overriding the `stroke` color on the parent `<g>` element in the update phase, so for me all circles are black. Changing to
```
.style('fill', 'none')
```
seems to fix it.

PR #322, opened 2019-10-07
use group color in clean up table

sgratzl:
based on #319 

apply group coloring in the cleanup table:

image


jeffbaumes:
Table text should toggle between light and dark based on the lightness value of the group color:

image

sgratzl:
image

PR #329, opened 2019-10-08
Limit redraw

subdavis:


web/src/components/vis/Heatmap.vue
```diff
@@ -383,9 +385,13 @@ export default {
       const i = Math.floor(evt.offsetY / h);
       const rnode = this.rowLeaves[i].data;
       const cnode = this.columnLeaves[j].data;
-      this.row.hovered = new Set(rnode.indices);
-      this.column.hovered = new Set(cnode.indices);
-      canvas.title = `${rnode.name} x ${cnode.name} = ${aggregate(this.values.data, rnode.indices, cnode.indices)}`;
+      if (rnode !== this.rnode || cnode !== this.cnode) {
```
sgratzl: `this.rnode` and `this.cnode` aren't updated by the other way to hover indices (via tree). Thus might be an inconsistent state. 

what about (pseudo code)
```js
function setHovered(wrapper, newValue) {
 if (wrapper.hovered.size === newValue.size && Array.from(newValue).every(d => oldValue.has(d)) {
   return false; // same
  }
  wrapper.hovered = newValue;
  return true;
}

const rowChanged = setHovered(this.row, new Set(rnode.indices));
const columnChanged = setHovered(this.column, new Set(cnode.indices));
if (rowChanged || columnChanged) {
  canvas.title = `${rnode.name} x ${cnode.name} = ${aggregate(this.values.data, rnode.indices, cnode.indices)}`;
}
```

subdavis: > this.rnode and this.cnode aren't updated by the other way to hover indices (via tree).

That isn't really a problem.  It will only cause a single extra render when moving between tree hover and node hover because `this.rnode` and `this.cnode` will be out of date.

I'd prefer to keep the simpler changes since I think they are much more readable.  They're also easily identifiable as O(1).

sgratzl: but what if
 a. in the canvas you move over rowNode A
 b. in the tree get over rowNode B
 c. in the canvas you move over rowNode A again

imo rowNode B will still be highlighted

subdavis: Does https://github.com/girder/viime/pull/329/commits/8f2bde81584aa518af2c59c8adddeb3a15a3054b take care of that?

PR #330, opened 2019-10-09
Heatmap coloring and filter

sgratzl:
based on #308 AND #319 

Allow the user to
 * filter columns by selection
 * have the rows and columns color coded according to their group / selection state

image


jeffbaumes:
Selected metabolites option does not seem to be working for me. I get the wrong metabolite colors in the bottom row (should all be orange in this state). I believe the right metabolites are shown, just colored incorrectly. I got in this state after selecting some metabolites from the ANOVA tab, and the incorrect coloring persists after page reload.

image

sgratzl:
fixed

PR #334, opened 2019-10-09
Sgratzl/wilcoxonimprovments

sgratzl:
improves the metabolite table used by Wilcoxon and ANOVA

image

@jeffbaumes this is what I meant with sliders / filters in the column headers.


sgratzl:
image

Issue #328, opened 2019-10-08
Heatmap computation is wrong after transforms


subdavis:
## Steps to reproduce

Start with the sgratzl/heatmapfix2 branch (or master).  I tried both.

1. load data_original.csv
2. View the heatmap.
3. Navigate to transform table, and change some stuff.  
  a. I made Normalize = minmax, Transform = Log10
  b. Leave plenty of time between each interaction to prove to yourself this isn't a race condition.
4. View the heatmap again.  It has changed since 1.
5. Refresh the page, the heatmap has changed a third time.  

I don't know whether it was correct in 4 or 5, but I suspect 4 is wrong.  Looking at the network call history, everything looks correct to me.   

## Screenshot from 4 

![Screenshot from 2019-10-08 16-32-57](https://user-images.githubusercontent.com/4214172/66431087-55581b00-e9e9-11e9-9560-010aa4d0e2dd.png)

sgratzl:
the heatmap isn't computed on the server but just takes the client version of the validated table. Can you please verify that the boxplots remain the same?

subdavis:
> Can you please verify that the boxplots remain the same?

Boxplots change as expected when you change the transformation method.   I didn't realize only the dendrogram was computed on the server.

I can see now that the color squares are the only thing are wrong.  After changing a transformation method, the dendrogram recomputes properly and dones't change on page reload.

subdavis:
Here's an example.  You can see that the dendrogram is the same as the image above.

![Screenshot from 2019-10-08 16-59-50](https://user-images.githubusercontent.com/4214172/66432866-0e6c2480-e9ed-11e9-867f-46e227dd3e8b.png)

subdavis:
I didn't realize what you meant by the box plot until my walk home.  You're right, the box plot is also wrong either before or after reload!

Not sure what the source of this bug is anymore, but it could very well be the vuex store.

sgratzl:
as far as I understood right now:

1. one changes the normalization method, e.g. to `minmax`
1. the server stored the changed value in the `ValidatedMetaboliteTable` of the CSV File
1. the server returns the newly created `measurements` heatmap

so far so good. However, 

4. the user reloads the page
1. the page is calling `/id`
1. it returns a mix of the `ValidatedMetabliteTable` and the `CSVFile` having the right normalization settings
1.  the page calls `/id/validate`
[...]

PR #336, opened 2019-10-09
copy normalization arguments when validating

sgratzl:
addresses #328 

applies the quick fix for now by copying the old values


subdavis:
Going to defer to the wisdom of @jbeezley on this one.   Thanks for investigating this.

Issue #345, opened 2019-10-17
Add multi block data fusion integration algorithm


jeffbaumes:
This may require additional clarification from Lilian:

https://app.box.com/file/540545442590

PR #346, opened 2019-10-17
clean/multiblock integration

### LINKED ISSUE #344: Add clean data fusion integration algorithm
This may at first only result in questions for Lilian:

https://app.box.com/file/503026492306

sgratzl:
closes #344 and #345 

based on the original R scripts. 

notes
 * rebuild R container
 * clean: in the end the script generates a normalized PCA of each matrix and then does an inner join to combine them (`merge(...)`). Thus, I reuse the existing simple method but transform the measurements before
 * multi block: here a new combined measurement pca table is generated

### DIFF
--- a/devops/viime/R/merge.R
+++ b/devops/viime/R/merge.R
@@ -46,39 +46,15 @@ compute_clean_pca <- function(prefix, measurements) {
   MS.scores.variance
 }
 
-#' multi_block_pca_merge
+#' compute_multi_block
 #'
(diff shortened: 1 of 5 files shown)


sgratzl:
Lilian's scripts so far end with a plotting of the some data but not returning the combined table

sgratzl:
so far we don't have a possibility to call the docker container within tests (like on the ci)

PR #343, opened 2019-10-16
Sgratzl/sidebar

sgratzl:
closes #342 

image

PR #347, opened 2019-10-21
Download Table

sgratzl:
based on #343 

Adds another point to the main bar: `Download Data`. Allowing the user to download the validated table, a filtered version of it, a list of metabolites, or a list of samples in CSV format

image

### DIFF
--- a/viime/views.py
+++ b/viime/views.py
@@ -645,10 +645,40 @@ def set_scaling_method(validated_table):
 
 
 @csv_bp.route('/csv/<uuid:csv_id>/validate/download', methods=['GET'])
+@use_kwargs({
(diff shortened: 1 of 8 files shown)


jeffbaumes:
@sgratzl can we include group and metadata rows/columns to the downloaded table?

sgratzl:
> @sgratzl can we include group and metadata rows/columns to the downloaded table?

done

sgratzl:
image

will become

image

sgratzl:
> Looks like the group column is still missing.

the group is already encoded in the row column by color and the group label is seen in the sample filter? should I duplicate it, if yes, still the row color coding?

PR #358, opened 2019-10-23
Sgratzl/dockertest

sgratzl:


tests/conftest.py
```diff
@@ -71,3 +74,32 @@ def pathological_table(client):
...
+
+    app.config['OPENCPU_API_ROOT'] = 'http://localhost:%s/ocpu/library' % port
+
+    container.start()
+    sleep(5)
+
+    yield opencpu_request
+
+    container.stop()
+    container.remove()
```
jbeezley: I would rather avoid mixing build-time concerns with testing.  The tests should just assume the container is running at the URI configured.  If we want to add starting up the opencpu service with the test environment [tox-docker](https://github.com/tox-dev/tox-docker) is a viable option.

PR #368, opened 2019-10-24
Sgratzl/uifixes

### LINKED ISSUE #367: conflict of group and selection colors
atm both the selection (orange, blue) and the default group colors (orange, blue, ..) use the same color scale (d3 category10). Thus there is a conflict. Any easy fix would be rotate the group color scale such that orange and blue will be the last one of the 10 not the first ones

sgratzl:
closes #367 and reorders integration methods

I changed the selection to orange/grey and removed orange from the group color scale

image

### DIFF
--- a/viime/models.py
+++ b/viime/models.py
@@ -245,8 +245,8 @@ class CSVFile(db.Model):
             return
         levels = sorted([str(v) for v in groups.iloc[:, 0].unique()])
         # d3 scheme category 10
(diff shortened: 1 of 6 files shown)


sgratzl:
> blue/green looks a bit close to me

it is actually https://github.com/d3/d3-scale-chromatic#schemeCategory10 without the orange.

PR #371, opened 2019-10-24
Sgratzl/downloadcharts

### LINKED ISSUE #364: Download for correlation and heatmap

sgratzl:
closes #364 

screenshot: 
image

download:
Heatmap (16)

screenshot:
image

download:
Correlation Network (8)

### DIFF
--- a/web/src/components/vis/Heatmap.vue
+++ b/web/src/components/vis/Heatmap.vue
@@ -4,6 +4,7 @@ import { hierarchy, cluster } from 'd3-hierarchy';
 import { scaleSequential } from 'd3-scale';
 import { interpolateBlues } from 'd3-scale-chromatic';
 import { select, event } from 'd3-selection';
+import { svg2url } from '../../utils/exporter';
 
 function extent(arr) {
   let min = Number.POSITIVE_INFINITY;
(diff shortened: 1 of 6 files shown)


jeffbaumes:
Not sure if it is from this branch or another recent one, but I'm getting the following problem now. I get an empty Clean Up Table view and empty charts. Is this a valid issue or is something I'm doing out of sync (db schema? docker?)

image

sgratzl:
> Not sure if it is from this branch or another recent one, but I'm getting the following problem now. I get an empty Clean Up Table view and empty charts. Is this a valid issue or is something I'm doing out of sync (db schema? docker?)

this branch doesn't touch the files you get an error. Have you recently switched branches. Maybe the hot reloading is not perfect. At least I cannot reproduce the error locally

jeffbaumes:
Yeah, when I restarted the client and server builds just now, the error went away.

web/src/components/vis/Heatmap.vue
```diff
@@ -454,6 +455,102 @@ export default {
       this.rnode = null;
       this.cnode = null;
     },
+
+    generateImage() {
```
jeffbaumes: This code looks new (not moved from elsewhere). Does the same thing need to now be rendered in two different ways to support the interactive view and download? Will code need to change in two places to keep things looking the same?

sgratzl: the heatmap is not just a simple SVG like other plots but
 * two SVGs for the dendograms
 * a HTML canvas  for the matrix
 * two divs with for the labels

Thus, cannot be handled in the same way as other plots

PR #372, opened 2019-10-24
WIP Sgratzl/imputehints

### LINKED ISSUE #362: Move number of missing values to sidebar
Possibilities:
* Color cells by imputation type
* Display number of metabolite columns of each type

sgratzl:
closes #362 

start with computing and transferring the method used to impute from R to python.

### DIFF
--- a/devops/viime/R/impute.R
+++ b/devops/viime/R/impute.R
@@ -18,9 +18,12 @@
 #' @export
 imputation <- function(table, groups,
                        mnar="zero", mcar="random-forest",
-                       p_mnar=0.70, p_mcar=0.40) {
+                       p_mnar=0.70, p_mcar=0.40, add_info=FALSE) {
   table <- read.csv(table, row.names = 1)
   if (sum(colSums(is.na(table))) == 0) {
+    if (add_info) {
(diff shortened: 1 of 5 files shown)


migrations/versions/e6f315c1f2f6_.py
```diff
@@ -0,0 +1,30 @@
...
+# revision identifiers, used by Alembic.
+revision = 'e6f315c1f2f6'
+down_revision = '92846651ae9d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('validated_metabolite_table', sa.Column('imputation_info', sqlalchemy_utils.types.json.JSONType(), nullable=True))
```
jbeezley: Ideally we should add the imputation info to the existing rows in this database.  What are the implications on the client if this column is null on a table for which imputation was performed?

sgratzl: see 
https://github.com/girder/viime/blob/a80832d30b55b63944e6b514d2ca1995cbafdc4f/web/src/components/Impute.vue#L71-L77

I implemented the client to handle null values properly

PR #389, opened 2019-10-30
Sgratzl/metametadata

### LINKED ISSUE #323: specify meta data column type
allow to specify / distinguish sample meta data based on the value type. e.g. age -> numerical, gender -> categorical

### LINKED ISSUE #375: Color columns by dataset type
From Tom's email. We have a row with the dataset for each column, and it would be good to be able to color by this in all analysis views, but particularly the heatmap.

Instead of making the dataset number a special case, this could be enabled in general for metadata rows. I could see it being more intuitive for the file name or some other identifier ("serum") to show up in a legend instead of the 1, 2, etc. number.

@sgratzl Any thoughts on how to solve this generally?

### LINKED ISSUE #377: Box plot tab
Filter by selection

Option to have bar per group

sgratzl:
closes #375 , closes #323 

notes
 * add the option to add metadata and a subtype to a column definition. So far there is no UI for it but it just used by the data source meta data (run `flask db upgrade`)
 * separate and generalize the idea of a metabolite/sample filter/colorer. The options are automatically derived from the metadata/groups/selection
 * use these new filter/colorer in: heatmap, correlation network, Wilcoxon and Anova Table

examples:

image

image

image

image

closes #377 

image


jeffbaumes:
Functionally, this works as expected from some limited manual testing. There may be some UI refactoring to do to possibly simplify or arrange options, but in general I'm happy with the controls layout.

A summary of what I've learned from the code for others who may want to take look at this:

* I can see that there is a new `subtype` db field which marks a metadata rows or columns as 'categorical', 'numerical', or 'ordinal'.
* There is also a new `meta` field which can be used to specify information about values contained within the row/column and how they should be mapped to UI labels and colors. An example is:

```
'meta': dict(levels=[dict(name='DS1', label='test_csv_file1.csv', color='#8dd3c7'),
                     dict(name='DS2', label='test_csv_file2.csv', color='#ffffb3')])
```

* One possible simplification is to remove label and always assume that what is in the dataset is what is displayed in the UI. In this example case, we would just always store the full dataset name in the row itself instead of the shortened `'DS1'`. But this is a minor simplification, and there may be unforeseen uses of this value/label separation, so I'm ok with leaving this in.
* There are several components for managing filtering and coloring of metabolites and samples.
* There is a bunch of additional logic in various analysis views to support these color/filter controls.
* There is a new box plot component with cool split-by-group controls which Tom had requested.

sgratzl:
> I get these console warnings on a merged dataset when I click on ANOVA:

@jeffbaumes can you please verify that these issues are related to this branch? i.e. if they also occur in the master branch

sgratzl:
> `text-overflow: ellipsis` would be good for longer dataset names in both the Node Filter and Node Color. Hover to see full text would also be good if we go this route.

done

sgratzl:
> > I get these console warnings on a merged dataset when I click on ANOVA:
> 
> @jeffbaumes can you please verify that these issues are related to this branch? i.e. if they also occur in the master branch

see https://github.com/girder/viime/pull/403

sgratzl:
> At a slightly lower level I'm unclear on how these new fields are intended to be used. Is the idea to expose setting metadata subtype via the clean up table? What information goes in the meta column, what components read and write to it?

so far there is no UI for editing it. I can imagine that instead of a single meta data row/column type you have multiple one (categorical meta data, numerical meta data, string meta data). But this is out of scope for the upcoming release.

subdavis:
I'm having an issue with flickering on the box plot:

https://drive.google.com/file/d/1BFLsz7qmohuSGRTVvNWsEwKx632cpgKH/view?usp=sharing

subdavis:
Thanks, c1e3d11 fixed the jitter for me.

Screenshot from 2019-11-01 15-59-28

Without the dendrogram, I think some padding is needed.  It looks like the heatmap runs off the top and left.  The group key column to the right actually does run off the top.

sgratzl:
> Without the dendrogram, I think some padding is needed. It looks like the heatmap runs off the top and left. The group key column to the right actually does run off the top.

I didn't change this part of the heatmap logic, just how the input data is fed in. Thus, imo it should be fixed separately. Otherwise, this PR is going to be mix too many things together

sgratzl:
> > Without the dendrogram, I think some padding is needed. It looks like the heatmap runs off the top and left. The group key column to the right actually does run off the top.
> 
> I didn't change this part of the heatmap logic, just how the input data is fed in. Thus, imo it should be fixed separately. Otherwise, this PR is going to be mix too many things together

see https://github.com/girder/viime/pull/390/commits/a0660bd828842799869621ff50f894d1cd5dbe1c

web/src/components/toolbar/ColorerOption.vue
```diff
@@ -0,0 +1,67 @@
...
+    },
+  },
+};
+</script>
+
+<template lang="pug">
+div(v-if="hasOptions")
+  v-toolbar.darken-3(color="primary", dark, flat, dense, :card="false")
+    v-toolbar-title
+      slot(name=title) {{title}}
```
subdavis: Can this slot be removed?  I does not appear to ever get used.   Otherwise, please quote attribute value.

web/src/components/toolbar/ColorerOption.vue
```diff
@@ -0,0 +1,67 @@
...
+
+  v-card.mx-3(flat)
+    v-card-actions(style="display: block")
+      v-select.my-0(:value="value", v-if="showSelect",
+          hide-details, :disabled="disabled",
+          @change="$emit('input', $event)",
+          :items="options", item-text="name")
+      .my-0.option(v-for="o in filterOptions", :key="o.name", :title="o.name")
+        v-icon(:color="o.color") {{ $vuetify.icons.square }}
+        | {{ o.name }}
```
subdavis: I've been struggling with this part for a while.   The color legend looks almost exactly like the controls, so it's very easy to confuse between the checkboxes and this, which is non-interactive.  

I wonder if these colorer cards could be combined with the control boxes, since as far as I can tell the information is duplicated.  The only issue is that the color isn't available when the box is unchecked, which I think we could solve by either changing the color of the unchecked box or putting a color dot to the right of the control.  

Screenshot from 2019-11-11 08-34-36

I know that a future PR adds group selection to these boxes.  I think this could be rolled into the same control as well.

sgratzl: so far if you can filter by selection but color by e.g. data source. In these cases the need to be two separate controls. Yes, when both show the same attribute information is redundant.

subdavis: Thanks. This is a hard control, but I see the case you've described.  I wonder if @waxlamp has any ideas about how to either unify or create distinction between the two.

jeffbaumes: Can we move this discussion to an issue? Due to the deadline it's important to get major functionality merged.

PR #403, opened 2019-11-01
small bugs and improvments

sgratzl:
notes
 * removes keep-alive since it creates a lot of problems with parameterized views (e.g. problem bar issue from #389 
 * that allowed to properly nest the analysis tabs
 * fix menu navigation issues
 * more robust wilcoxon and anova plots

### DIFF
--- a/web/src/App.vue
+++ b/web/src/App.vue
@@ -20,7 +20,7 @@ export default {
           text: 'VIIME',
           to: { name: 'Root' },
         },
-        ...this.$route.matched.map((route) => {
+        ...this.$route.matched.filter(route => !route.meta.hidden).map((route) => {
           const b = route.meta.breadcrumb;
           return {
             text: route.name,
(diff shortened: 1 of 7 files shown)


web/src/App.vue
```diff
@@ -20,7 +20,7 @@ export default {
           text: 'VIIME',
           to: { name: 'Root' },
         },
-        ...this.$route.matched.map((route) => {
+        ...this.$route.matched.filter(route => !route.meta.hidden).map((route) => {
```
subdavis: For routes without the `meta` property, won't you get `cannot read property 'hidden' of undefined` ?

That case should probably be handled since `meta` isn't a sanctioned router property.

sgratzl: what do you mean, meta is a valid field: https://router.vuejs.org/guide/advanced/meta.html

subdavis: Oh, weird.  I've never noticed it.

Issue #387, opened 2019-10-30
Score plot group coloring


jeffbaumes:
Score plot dots should be react to changes in dataset group color selections. This was lost in the conversion to C3 in #219.

PR #405, opened 2019-11-01
Back out abuses of computed properties in chart components

waxlamp:
@subdavis, could you take a look at this and see if it's kosher? If yes, I'll apply the same strategy to the other chart components.

### DIFF
--- a/web/src/components/vis/ScorePlot.vue
+++ b/web/src/components/vis/ScorePlot.vue
@@ -194,7 +194,7 @@ export default {
         && eigenvalues.length > 0;
     },
 
-    update() {
+    updateDeps() {
       const {
         pcPoints,
         pcX,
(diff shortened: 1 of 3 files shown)


sgratzl:
can you elaborate why this solution is better than the one currently used?

web/src/components/vis/ScorePlot.vue
```diff
@@ -208,111 +208,24 @@ export default {
...
-      //
-      // In general, this is part of the tradeoff of mixing an imperative vis
-      // library like C3 with a reactive framework like Vue. The alternatives are
-      // less attractive than breaking the rules here: for instance, we could set a
-      // series of watchers on the dependencies of this function to explicitly
-      // trigger it, but that presents a serious danger of falling out of sync with
-      // changes in the dependency list.
-      //
-      // eslint-disable-next-line vue/no-side-effects-in-computed-properties
-      this.labels = labels;
```
subdavis: @sgratzl this is the main reason.  Mutations inside of computed properties is discouraged.  Mutations inside watchers is perfectly acceptable.

sgratzl: I mean not the ScorePlot but other plots that are using a slightly different approach such as ForceDirectedGraph, ....

sgratzl: as far as I can, with this change I separate the data needed by a plot and the rendering. Thus, if one uses by accident `this.xxx` inside the `update` method directly, it won't be tracked.

sgratzl: Moreover, in this specific case, why is `this.labels` not a computed property directly? Moreover, this logic will change once #387 is implemented again.

subdavis: That's a fair point. In my experience, location is one of the things I'm most aware of when writing vue code.  I think this pattern is more idiomatic Vue because even a novice Vue user would not expect a method to get called in the same way a computed prop is.

I'm not sure if that makes sense.

waxlamp: @sgratzl I hadn't realized you'd put a very similar mechanism into the `axisPlot` mixin. After the November release, I think it would make a lot of sense to abstract that functionality into its own mixin, and use it in *any* component that relies on imperative, non-Vue mechanisms to fulfill their role.

I will close this PR and open an issue to tackle this later. Thanks for weighing in on this.

waxlamp: @subdavis pointed out to me off-channel that there may still be some merit to this PR. I'll leave it open but we can consider the issues more fully after the November release.

jeffbaumes:
@waxlamp is this ready to go after resolving merge conflicts?

waxlamp:
> @waxlamp is this ready to go after resolving merge conflicts?

I don't believe so--I would mark this back to "draft" if GitHub allowed that.

This PR is more of an Exhibit A for a discussion about how to properly mix Vue's reactivity with our vis libraries' imperativity.

jeffbaumes:
@waxlamp should this be merged or abandoned? There are now conflicts on some of the files.

subdavis:
I still think it has value.  Conflicts are my fault, so I could resolve those if @waxlamp is busy.  I don't think any of these changes conflict with any other ongoing efforts.

waxlamp:
> I still think it has value. Conflicts are my fault, so I could resolve those if @waxlamp is busy. I don't think any of these changes conflict with any other ongoing efforts.

Please go ahead and resolve conflicts and get this ready for merge as you see fit (I've made you a co-assignee of the PR). I think you have a better handle on this PR at this point than I do.

I will offer one stray thought: it seems as though this could benefit greatly from the "multiple watcher" mechanism afforded by composition API.

jeffbaumes:
Have been resolved with recent code updates.

PR #414, opened 2019-11-04
small upload fixes

### LINKED ISSUE #198: Data source meta data

### LINKED ISSUE #314: Uploading the same dataset twice
In the overview when I upload a file using the file chooser and then try to upload the same file again using the file chooser -> nothing happens. 

possible reason: the `input[type=file]` won't be reseted after the upload thus when selecting the same file again, no `change` event is fired.

sgratzl:
small fixes in the upload page

closes #314, closes #198

image

### DIFF
--- a/web/src/components/Upload.vue
+++ b/web/src/components/Upload.vue
@@ -64,6 +64,7 @@ export default {
       return Object.keys(datasets).map((id) => {
         const d = datasets[id];
         return {
+          id: d.id,
           file: {
             name: d.name,
             size: d.size, // TODO: fix when server implements this.
@@ -75,7 +76,7 @@ export default {
       });
     },
     files() {
-      return this.readyFiles.concat(this.pendingFiles.filter(f => f.status !== 'done'));
+      return this.readyFiles.concat(this.pendingFiles);
     },
     deleteDialog() {
       return this.deleteCount > 0;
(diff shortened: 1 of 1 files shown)


web/src/components/Upload.vue
```diff
@@ -6,22 +6,6 @@ import FileList from '@girder/components/src/components/Presentation/FileUploadL
...
-];
-
-const dataTypes = [
-  { name: 'NMR Concentrations', value: 'nmr' },
-  { name: 'conc', value: 'conc' },
-  { name: 'GC/MS', value: 'gcms' },
-  { name: 'LC/MS', value: 'lcms' },
-  { name: 'Other', value: 'other' },
-];
-
```
subdavis: Is this a temporary removal until we can add backend support?  Or did Tom change his mind about wanting this?  Just curious.

jeffbaumes: This is unclear. As far as I am aware it has no analytic value, so this type of information can always be placed in the name/description. I think it's good to remove for now and put it back if/when we know it's needed.

sgratzl: in the end we always have the git history

PR #415, opened 2019-11-04
start adding favicon

sgratzl:
using https://realfavicongenerator.net/

image

@jtomeck you know better regarding spacing


jtomeck:
@sgratzl It doesn't look a ton different than the images you used, but the logo file I worked from had some slight changes that I wanted to make sure got updated, so I replaced them.

@jeffbaumes should we discuss the plan for the breadcrumbs, or do you think that should live in a different PR?

sgratzl:
btw. the rich diff feature of Github is really nice:

https://github.com/girder/viime/pull/415/commits/8786fb89780050d4765d2ea5a343bac58188abd7?short_path=f944fde#diff-f944fdef49d8bfd3e1cfb3b92664fe7d

sgratzl:
> @jeffbaumes should we discuss the plan for the breadcrumbs, or do you think that should live in a different PR?

I would say a different one, since changing the breadcrumbs requires the landing page while this PR is independent of it.

PR #426, opened 2019-11-08
tune external links

sgratzl:
use open-in-new icon since it links out to an external page + set proper target and rel attributes

image


subdavis:
The use of `target="_blank"` is heavily debated, but I'm of the opinion that it should almost never be used.  Browsers provide the "open in new tab" right click option for users who want this.  

https://css-tricks.com/use-target_blank/ provides a pretty good explanation of my position.

That said, there's plenty of people who disagree.  I'll defer to UX guru @jtomeck on this one.

PR #429, opened 2019-11-08
Mask image for banner

jtomeck:
Here is the SVG of the image to do the masking for.


sgratzl:
new version using dummy image (original is already masked)

image

jtomeck:
I just tried swapping the image, but could not get it working. In the interest of time, I'll just post the images here and defer to your implementation @sgratzl if that's ok.

viime_screen_lg
Large version

viime_screen_sm
Small version

sgratzl:
done (even with nice srcset for optimized image loading)

image

jeffbaumes:
In the mobile layout, it seems the image stretches. Here it is around 850px wide:

image

PR #439, opened 2019-11-11
help texts

sgratzl:
add a bunch of help texts based on the google doc (analysis methods, correlation method, impute, merge methods)

e.g. 
image

image


subdavis:
If we ever want to do internationalization, some things are harder to change than others.   Text inline in paragraphs of a template is easy enough to swap, but putting raw html into json documents may prove harder to back out, particularly if that sets a precedent that future contributors follow.

That's probably not a best practice anyway since it's not possible to parse/lint those items.

Even if we don't intend to translate, https://kazupon.github.io/vue-i18n/guide/interpolation.html#basic-usage provides some nice tools for dealing with formatted interpolation.

Just something to consider.

sgratzl:
is there a way in vue to dynamically render components / templates? what I want is:

```
items = [{
 name: 'a', 
 desc: ???
}]
...
div(v-for="i in items", :key="i.name")
  .name {{i.name}}
  .desc ??? render i.text as as vue HTML ???
```

In react I would just do it:
```
items = [{
 name: 'a', 
 desc: <p>Help Text</p>
}]
...
{items.map((item) => (
  <div key={item.name}>
    <div className="name">{item.name}</div>
    <div className="desc">{item.desc}</div>
  </div>)
}
```

subdavis:
Not sure I understand.  The example you provided looks the same to me as the code in this PR: you have html inside a json object that won't be linted.

Vue has "dynamic components" https://vuejs.org/v2/guide/components-dynamic-async.html but I don't think that's the same concept you're looking for.   I'm not sure what the value of `desc: ???` is supposed to be.

If you're talking about the `<code>` tags, I'm not sure how to solve that problem...

sgratzl:
nevermind I figured it out based on https://vuejs.org/v2/guide/render-function.html

Issue #437, opened 2019-11-11
Test on Edge browser


jeffbaumes:
Seems to work ok on Chrome/FF.

I think Microsoft has some free VMs for testing out their browsers.

We may also want to see what happens on IE 11.

subdavis:
I'll take a look at this today.

subdavis:
* Download png doens't work anywhere.  No errors in console
* Download table works, but metabolite list and sample list do not.

subdavis:
Up and down arrows do not appear next to these controls anywhere in the app

![Capture4](https://user-images.githubusercontent.com/4214172/68684680-e2394b80-0536-11ea-8853-2b6c5f248118.PNG)

subdavis:
Select styling is wrong

![Capture7](https://user-images.githubusercontent.com/4214172/68684719-f54c1b80-0536-11ea-86fc-5be35dd629bb.PNG)

subdavis:
PCA plot ellipse styling is broken

![Capture3](https://user-images.githubusercontent.com/4214172/68684740-ff6e1a00-0536-11ea-936d-bc41dd50308f.PNG)

subdavis:
* Horizontal scroll bar in data validation issue navbar.  I started over and could not reproduce, so this is just something to look out for.

![Capture6](https://user-images.githubusercontent.com/4214172/68684807-16147100-0537-11ea-9868-5f2c25e5209d.PNG)

subdavis:
Landing page scaling issues.  cc @jtomeck 

![Capture](https://user-images.githubusercontent.com/4214172/68685154-ad79c400-0537-11ea-866d-a9ec2355ed15.PNG)

subdavis:
Vue infinite scroll table becomes confused.  This appears to be related to the uncommon browser behavior where scrolling "past" a page's end causes it to stretch past and snap back into alignment when you let go. I believe Safari does the same thing.  

I don't think we can really do anything about this.  you sort of have to "shake" the table by scrolling around fast to get into this state.

/shrug

![Capture2](https://user-images.githubusercontent.com/4214172/68685242-d13d0a00-0537-11ea-9720-9870d0929521.PNG)

jeffbaumes:
With this many issues, @subdavis do you think it would be best to do browser detection and put a blocking warning on the front page if using IE/Edge saying it is not supported and to use Chrome/FF/Safari (we still need to check Safari)? Or @subdavis @sgratzl do you think these are resolvable in the next day or two?

[...]

PR #458, opened 2019-11-12
Add browser support banner

subdavis:
Didn't really want to add an external dependency.  Found the best implementation on S/O and dropped it into a mixin.  I agree with https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser that this is better than user agent parsing.

Going to leave #437 open because this doesn't really resolve any bugs.

Screenshot from 2019-11-12 13-59-06


subdavis:
I actually think the snackbar looks better:

Screenshot from 2019-11-12 14-44-40

But it gets tricky and annoying on mobile, and I'd have to add a lot of extra custom styling to fix it, so I think I'll leave the simpler toolbar banner.

Screenshot from 2019-11-12 14-46-02

PR #460, opened 2019-11-12
tune heatmap

sgratzl:
tune heatmap rendering performance. 

at least on my machine, a mousemove operation took around 120ms before and now around 40ms

### DIFF
--- a/web/src/components/vis/Heatmap.vue
+++ b/web/src/components/vis/Heatmap.vue
@@ -47,9 +47,9 @@ function aggregate(arr, rs, cs, transposed) {
 const DENDOGRAM_RATIO = 0.2;
 const LABEL_WIDTH = 150;
 
-const MDI_PLUS_CIRCLE = '&#xF417;';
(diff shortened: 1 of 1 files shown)


web/src/components/vis/Heatmap.vue
```diff
@@ -47,16 +47,38 @@ function aggregate(arr, rs, cs, transposed) {
...
+    return false;
+  }
+
+  const old = wrapper.hovered;
+
+  if (old.size !== arr.length) {
+    wrapper.hovered = new Set(arr);
+    return true;
+  }
+  if (old.size === 0 || ((arr.length === 1 && old.has(arr[0])) || arr.every(d => old.has(d)))) {
```
subdavis: Could you add a comment about what this does?

sgratzl: just compares the array with the set with optimizations for 0 length and length 1

subdavis: Please add some documentation on this function.

Issue #320, opened 2019-10-07
Node-link rendering performance


jeffbaumes:
@subdavis from #299:

FWIW, with the 1000 metabolite file, rendering the node graph still causes the UI to freeze for several seconds and each render loop appears to need about 150-200 MB of memory with default settings. Memory usage scales pretty linearly with how many nodes are on screen (expected).

Resizing in particular takes a while because a sequence of 2-3 second operations happens before you get to the current resize.

I poked around a bit and https://blog.scottlogic.com/2017/10/30/migrating-d3-force-layout-to-webassembly.html might be worth exploring.

sgratzl:
regarding freezing: a quick win is to refactor:

https://github.com/girder/viime/blob/27fb23da4e108b93bee74d7435c16813b6ac9d63/web/src/components/vis/ForceDirectedGraph.vue#L139

in multiple steps where each step is e.g. using https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback or another way (`setTimeout`, `setInterval`) to allow the browser to do something in between. If that is still too slow, we can extract the force computation to a Webworker.

PR #461, opened 2019-11-12
use idlecallback for force batch computation

sgratzl:


jeffbaumes:
@sgratzl Could you briefly explain what this PR resolves?

sgratzl:
it implements: https://github.com/girder/viime/issues/320#issuecomment-539115604

PR #465, opened 2019-11-13
Sample dialog

jtomeck:
Took a stab at making this look a bit better. I did not have sample data in my local copy, so I was sort of styling blind, but I think this should help, at least.

Screen Shot 2019-11-13 at 11 34 56 AM

### DIFF
--- a/web/src/components/SampleBrowser.vue
+++ b/web/src/components/SampleBrowser.vue
@@ -67,11 +67,13 @@ export default {
 .import {
   flex: 0 0 auto;
 }
+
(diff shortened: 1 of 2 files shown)


sgratzl:
fixed some styles: now looks locally like:

image

jeffbaumes:
I'll merge since @jtomeck might not be following all notifications.

jtomeck:
> I'll merge since @jtomeck might not be following all notifications.

Thanks Jeff, just catching up on everything now.

PR #473, opened 2019-11-13
Sgratzl/numericgroups

### LINKED ISSUE #469: Groups not always detected correctly
Uploading this file:

[data100.xlsx](https://github.com/girder/viime/files/3842902/data100.xlsx)

Gives this state (note that groups are not identified and colored):

image

It may be the numeric group column?

sgratzl:
closes #469 

image

two relevant changes:
 1. enforces that the group data frame is a string
 2. show only values (sample x measurement) formatted otherwise show the value as is


jeffbaumes:
@sgratzl ready to merge.

PR #474, opened 2019-11-13
loading scale rescale (method 4)

### LINKED ISSUE #472: loadings plot scale
so far the loading plot scales are fixed to -1 ... 1 regardless of the data based on the assumption that correlations values are just between -1 and 1. However, with only small values in the data, the plot is a dot in the center. 

possible axis scales:
 1. each axis to min/max of the actual values
 1. each axis to `-Math.max(abs(min), abs(max))` to `Math.max(abs(min), abs(max))` .. symmetric
 1. both axes to min/max of either of those
 1. both axis to symmetric of either of them

sgratzl:
implements the fourth method as described in #472 

closes #472

### DIFF
--- a/web/src/components/vis/LoadingsPlot.vue
+++ b/web/src/components/vis/LoadingsPlot.vue
@@ -45,7 +45,6 @@ function domain(arr) {
       max = v;
     }
   });
-  // return [min, max];
   const amax = Math.max(Math.abs(max), Math.abs(min));
   return [-amax, amax];
 }
@@ -118,11 +117,9 @@ export default {
       return [Math.min(x[0], y[0]), Math.max(x[1], y[1])];
     },
     xrange() {
-      // return [-1, 1];
       return this.pcRange;
     },
     yrange() {
-      // return [-1, 1];
       return this.pcRange;
     },
   },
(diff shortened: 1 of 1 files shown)


web/src/components/vis/LoadingsPlot.vue
```diff
@@ -74,6 +90,41 @@ export default {
...
+      }
+      return domain(arr);
+    },
+    pcRange() {
+      const x = this.pcXRange;
+      const y = this.pcYRange;
+      return [Math.min(x[0], y[0]), Math.max(x[1], y[1])];
+    },
+    xrange() {
+      // return [-1, 1];
```
waxlamp: Please remove this line.

web/src/components/vis/LoadingsPlot.vue
```diff
@@ -74,6 +90,41 @@ export default {
...
+      return this.pcRange;
+    },
+    yrange() {
+      // return [-1, 1];
```
waxlamp: Please remove this line.

waxlamp:
> @waxlamp Tom specifically requested this change, so we have his buy-in. He would rather see the noise than see a tight indecipherable blob in the center.
> 

Sounds good. Once more, I miss things by not being here :nerd_face: 

> @sgratzl I agree there were several choices that could have been made on how to set the bounds. I think this approach is reasonable.

I didn't mention this in my review, but I agree with this statement.

waxlamp:
Thanks for this, @sgratzl.

PR #477, opened 2019-11-14
Sgratzl/landingtune

sgratzl:
image

 * capital I in Visualization
 * bigger 'scroll to learn more..'


waxlamp:
I personally think we should leave the I lowercase.

sgratzl:
> I personally think we should leave the I lowercase.

see discussion yesterday in the meeting and the latest poster version. The idea was to emphasize the VIIME acronym.

waxlamp:
Gotcha, didn't realize there was a larger discussion yesterday. Thanks!

PR #479, opened 2019-11-14
Sgratzl/sampleorder

sgratzl:
allow to add an order column for sample groups to enforce a order. By default by name. In addition, sort sample files by name.


sgratzl:
> @sgratzl when this is in can you ensure that the "set of 3" dataset is listed first on viime.org?

yes and done

PR #481, opened 2019-11-14
Ensure score plot only displays currently selected dataset

### LINKED ISSUE #478: score plot union of data
having two dataset with different groups, e.g: 

image

and: 
image

however, when first viewing the first dataset and then switch to the other without reloading the page it looks like:

image

like that both datasets are encoded in the same plot (see legend)

waxlamp:
Closes #478

### DIFF
--- a/web/src/components/vis/ScorePlot.vue
+++ b/web/src/components/vis/ScorePlot.vue
@@ -284,6 +284,7 @@ export default {
       this.labels = labels;
 
       // Draw the C3 chart.
+      this.chart.unload();
       this.chart.load({
         columns,
         xs,


sgratzl:
have you read this note in the c3 docs:

https://c3js.org/reference.html#api-unload

> Note:
> If you call load API soon after/before unload, unload param of load should be used. Otherwise chart will not be rendered properly because of cancel of animation.

waxlamp:
Unfortunately, using the `unload` param of the `load()` method results in a weird flashing of the plot. It seems to be unloading and reloading data in an infinite loop. This may have to do with our use of `update()` as a computed property, but I can't be sure.

For now, I'd recommend doing this the unsafe way, then revisiting this issue as part of a push to migrate `update()` out of the computed property section.

PR #488, opened 2019-11-19
TypeScript v2

sgratzl:
another try to slowly migrate to typescript

missing
 * typing vuex e.g. using https://championswimmer.in/vuex-module-decorators/


jeffbaumes:
Even though this is still draft, I'd like @waxlamp and @subdavis to review the approach and comment with thoughts sometime before Thanksgiving.

web/src/components/toolbar/ColorerOption.vue
```diff
@@ -1,44 +1,54 @@
...
+  @Prop({
+    default: '',
+  })
+  readonly title!: string;
+
+  @Prop({
+    default: null,
+    type: String,
+  })
+  readonly value!: string|null;
```
waxlamp: Does the `!` tell TSC that this value will not be null or undefined? If so, how does that square with the specified type (and default value)?

sgratzl: generally yes, in this case it will tell typescript that the property will be initialized by someone

sgratzl: usually I would define a default value. However, it is not allowed to initialize a property in that way

waxlamp:
@jeffbaumes, what do you see as the path forward on this? Changes like these become difficult the longer we wait to merge them.

web/declarations.d.ts
```diff
@@ -0,0 +1,14 @@
...
+  export default content;
+}
+
+
+declare module '!url-loader*' {
+  const content: string;
+  export default content;
+}
+
+declare module 'vue-resize-directive';
```
subdavis: What's going on in this file?

web/src/components/FeedbackButton.vue
```diff
@@ -1,25 +1,25 @@
...
+  })
+  readonly front!: boolean;
+
+  visible = false;
+
+  readonly form = 'https://docs.google.com/forms/d/e/1FAIpQLScWccZqTF_b6nQkfkAvLNf7XGV9k7Rqf8V_Mid-oZLY47g_IQ/viewform';
+
+  readonly width = 640;
+
+  readonly height = 930;
```
subdavis: Is there a different way to express the data block?  This looks like how I express non-reactive properties in other places.  Might get used to it though.  Just curious.

waxlamp: This is one place where I find the Vue.extend() way to be better (though that is due mainly to familiarity with it). Here's an [example from Multinet](https://github.com/multinet-app/multinet/blob/master/client/src/views/GraphDetail.vue#L204-L212). The `as` keyword has a natural idiomatic usage here in explicitly providing types for data props.

subdavis: Yeah I dig that syntax.  Ah well, might just be one of those small concessions.

subdavis:
Talked with @roni and we identified some questions that would be nice to answer before accepting the class-style.

* difference of type-safety features between the two: is there an example project or some docs somewhere that explains this?  Should we create our own comparison demo?
* Different editor feature support.  I use VS Code and I care about those considerations.  Vetur has a lot of new features coming out, so we should take a look at what exactly the gaps are between the styles.
* Docs support and ability to onboard new team members.   I think there's a lot of value in being consistent with Vue.js conventions such that anybody familiar with Vue could jump in and feel comfortable.  @waxlamp also pointed out that the documentation for class-style is definitely lacking compared to object-style, which is an important consideration.  Part of Vue's value is the excellence of its docs, and we'd be sacrificing a lot here.
* There may be Vue 3.0 to consider as well, though we're still stuck on Vuetify 1.5 and that will be a painful upgrade, so an upgrade strategy may be important as we talk about this.


These changes may be worth the cost, but I'd like to know more specifics about what we gain and lose by making this choice.  
Thanks everyone.

waxlamp:
> Talked with @waxlamp and we identified some questions that would be nice to answer before accepting the class-style.
> 
> * difference of type-safety features between the two: is there an example project or some docs somewhere that explains this?  Should we create our own comparison demo?
> * Different editor feature support.  I use VS Code and I care about those considerations.  Vetur has a lot of new features coming out, so we should take a look at what exactly the gaps are between the styles.
> * Docs support and ability to onboard new team members.   I think there's a lot of value in being consistent with Vue.js conventions such that anybody familiar with Vue could jump in and feel comfortable.  @waxlamp also pointed out that the documentation for class-style is definitely lacking compared to object-style, which is an important consideration.  Part of Vue's value is the excellence of its docs, and we'd be sacrificing a lot here.
> * There may be Vue 3.0 to consider as well, though we're still stuck on Vuetify 1.5 and that will be a painful upgrade, so an upgrade strategy may be important as we talk about this.
> 
> These changes may be worth the cost, but I'd like to know more specifics about what we gain and lose by making this choice.
> Thanks everyone.

The tech meeting we are supposed to have next week would be a great time to discuss some of these questions.

web/declarations.d.ts
sgratzl: there are no typing information for libraries like `vue-resize-directive`. `.d.ts` files are definition files to tell TypeScript what the types of certain libraries are.

web/src/components/FeedbackButton.vue
sgratzl: in this case the typings are not needed since for `width = 640` Typescript knows that is a number. for an array I would type `arr: number[] = [];`

sgratzl:
btw. I'm currently going nuts when it comes to typings in https://github.com/Kitware/UPennContrast/

PR #493, opened 2019-11-20
closes #490 log2 transform

### LINKED ISSUE #490: log2 transform

sgratzl:


sgratzl:
> Is there a reason we're not using `np.log10()` and `np.log2()` directly?

I just used the log10 case as basis

PR #494, opened 2019-11-20
Fix score plot wonkiness

### LINKED ISSUE #484: Reactivity broken on Transform Table
#481 broke reactivity on the score plot on the Transform Table page when changing transform options. Oddly, it seems the ellipses may be going to the right place but points are not. Backing out the 1-line change restores working functionality. Seems ensuring the score plot updates when switching datasets will need to be fixed a different way.

waxlamp:
This PR performs a number of code transformations in order to eliminate some of the weird chart behaviors we've been seeing:
1. It moves the `update()` function out of computed properties and into a method. This seems to eliminate strange issues resulting from our abuse of Vue's reactivity system.
2. It performs a conditional `unload` of C3's data, depending on if the column names have changed (as happens specifically when the dataset is changed in Viime).
3. It computes the SVG transforms for confidence ellipses using the D3 scale objects, rather than chaining separate transforms together (resulting in simpler code and simpler transforms in the end).
4. It makes proper use of C3's load/unload API to ensure that the scale objects used in (3) above are the correct ones.

The PR introduces a new code pattern. Since we want `update()` called reactively on its dependencies, I have abstracted the dependencies themselves out into a computed property, set a watcher on that property that simply calls `update()`, and I've made sure to write `update()` such that it *only* takes props from the computed dependency property, and nowhere else. This is one of the tradeoffs needed in structuring the code this way (it replaces the earlier tradeoff of abusing the reactivity system).

Closes #484.
Depends on #486.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -56,7 +56,7 @@
     "eslint": "^5.8.0",
     "eslint-plugin-vue": "^5.0.0",
     "git-revision-webpack-plugin": "^3.0.3",
-    "node-sass": "^4.11.0",
+    "node-sass": "^4.13.0",
     "pug": "^2.0.3",
     "pug-lint-vue": "^0.1.3",
     "pug-plain-loader": "^1.0.0",
--- a/web/src/components/vis/ScorePlot.vue
+++ b/web/src/components/vis/ScorePlot.vue
@@ -482,18 +482,6 @@ export default {
         xs[g] = xName;
       });
 
-      // The following side effect depends on xData and yData in the same way
-      // that the C3 chart itself does, so we disable the linter warning that would
-      // prevent it.
-      //
-      // In general, this is part of the tradeoff of mixing an imperative vis
-      // library like C3 with a reactive framework like Vue. The alternatives are
-      // less attractive than breaking the rules here: for instance, we could set a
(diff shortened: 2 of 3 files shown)


subdavis:
@waxlamp this looks like it conflicts with #405 -- can that PR be closed?

web/package.json
```diff
@@ -56,7 +56,7 @@
     "eslint": "^5.8.0",
     "eslint-plugin-vue": "^5.0.0",
     "git-revision-webpack-plugin": "^3.0.3",
-    "node-sass": "^4.11.0",
```
subdavis: Why?

waxlamp: *sigh*

Because I upgraded to Node 13, which won't work with node-sass 4.11, and it's difficult to downgrade Node to a version that works with that version of node-sass.

This commit is actually in a separate PR: https://github.com/girder/viime/pull/486.

subdavis: Tested and still working with node LTS 10.16 so this is fine with me.

FWIW, https://github.com/tj/n is a thing that makes my life easier.

waxlamp:
> @waxlamp this looks like it conflicts with #405 -- can that PR be closed?

Yes we can close that, though we should still have the discussion generated from it.

PR #496, opened 2019-11-28
VolcanoPlot

sgratzl:
first version of a volcano plot for wilcoxon and anova:

image

### DIFF
--- a/devops/viime/R/analyses.R
+++ b/devops/viime/R/analyses.R
@@ -2,7 +2,7 @@
 #' wilcoxon_test_z_scores
 #'
 #' @export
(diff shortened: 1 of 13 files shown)


jeffbaumes:
Add

* Slider for significant fold change
* Slider for significant p-value
* 1 versus the rest with ANOVA in addition to choosing a pair
* Investigate color by, including by meeting each threshold (or both)

jeffbaumes:
@waxlamp please review by Mar 13 if possible, thanks.

Here is the current status. I split out "1 vs rest" into a separate issue #504. 

volcano

PR #498, opened 2019-12-02
lazy import certain modules for better module splitting

sgratzl:
lazy load C3 related chunks and just use the parts of lodash that are needed. Moreover, load the exported needed fonts on the fly. 

before:
image

after:
image

like the initial vendor bundle to load is reduced by 600kb or 33% and will be lazy loaded thus the landing page should appear earlier

to see it yourself:

```sh
cd web
yarn build --report-json
npx webpack-bundle-analyzer ./dist/report.json
```


subdavis:
This may be a good change, but I think it's worth a discussion.

Our API calls are still hundreds of milliseconds on average.  The Anova analysis takes 2 full seconds to return.  The UI has a number of usability problems that stem from timing issues that will require some thoughtful consideration.

I checked the timings on the live site from an empty cache and it looks like 50-100ms is the average.  From a user's perspective, I don't think anyone is going to care (or notice) that our load time improves by 20-40ms if everything else is taking so much longer.  This is potentially a premature optimization.

Thoughts?

sgratzl:
my main use case was that when the user first visits the landing page it should show up fast. Once they play around with the app it is a different story

jeffbaumes:
It was decided this is not worth implementing for now.

PR #502, opened 2020-01-10
Enable google analytics in production

jbeezley:


jeffbaumes:
Do we want the id directly in the code, or should we pull it from an env variable that is set in CI?

jbeezley:
I don't know what the standard is for this.  It's not a secret value because it is served with the app.  I think the only reason to move the value to CI would be if we had multiple deployments of the same code.

jeffbaumes:
Agree that it's not a security thing. But since we're open source, seems like a friendly move to not have all deploys send to our GA account.

PR #509, opened 2020-03-25
Implement get_row_by_name on CSVFile

jbeezley:
This is called by the table validation methods and is the cause of sentry issue [1559794771](https://sentry.io/organizations/kitware-data/issues/1559794771).


waxlamp:
How are there two PRs opened for the same branch? :open_mouth:

PR #511, opened 2020-03-26
Add a switch to turn off heatmap's "dendrogram mode"

waxlamp:
This adds a switch to the "Dendrogram" control panel in the heatmap tile, which will use the original order of the metabolites in the heatmap. This was requested by Tom, as sometimes the original order is meaningful on its own. Previously, the presence of the dendrogram meant that clustering was always performed and this original ordering was lost.

### DIFF
--- a/web/src/components/vis/HeatmapTile.vue
+++ b/web/src/components/vis/HeatmapTile.vue
@@ -127,7 +127,7 @@ vis-tile-large(v-if="plot", title="Heatmap", expanded, download, :download-impl=
     sample-colorer(:dataset="dataset", v-model="sampleColor")
 
     v-toolbar.darken-3(color="primary", dark, flat, dense)
-      v-toolbar-title.switch-title Dendrogram
+      v-toolbar-title Dendrogram
     v-card.mx-3(flat)
       v-card-actions(:style="{display: 'block'}")
         v-checkbox.my-0(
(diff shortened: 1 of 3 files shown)


jeffbaumes:
I see the request and should be able to review this today.

waxlamp:
Tom asked for a bit more fine-grained control over how this works but I think I'd prefer to get this one approved and merged and then work on followup features in new branches.

web/src/components/vis/HeatmapTile.vue
```diff
@@ -118,19 +127,41 @@ vis-tile-large(v-if="plot", title="Heatmap", expanded, download, :download-impl=
...
+  display: flex;
+  justify-content: space-between;
+  overflow: visible;
+  width: 100%;
+  .switch {
+    flex-grow: 0;
+    margin-right: -10px;
+  }
+}
+</style>
```
jeffbaumes, line 167: Since you removed this switch, should this CSS be removed?

web/src/components/vis/HeatmapTile.vue
```diff
@@ -118,19 +127,41 @@ vis-tile-large(v-if="plot", title="Heatmap", expanded, download, :download-impl=
     sample-colorer(:dataset="dataset", v-model="sampleColor")
 
     v-toolbar.darken-3(color="primary", dark, flat, dense)
-      v-toolbar-title Dendogram
+      v-toolbar-title.switch-title Dendrogram
```
jeffbaumes, line 130: Should the `.switch-title` class be removed?

web/src/components/vis/HeatmapTile.vue
waxlamp: Good catch--see 0030e51.

web/src/components/vis/HeatmapTile.vue
waxlamp: See 0030e51.

PR #516, opened 2020-05-26
Fix horizontal scrollbar at bottom of ANOVA table

mvandenburgh:


web/src/components/vis/AnovaTableTile.vue
```diff
@@ -89,10 +89,14 @@ vis-tile-large(v-if="plot", title="Anova Table", :loading="plot.loading", expand
     v-btn(flat, dark, block, @click="downloadTable")
       v-icon.mr-2 {{ $vuetify.icons.save }}
       | Download Table
-  anova-table(v-if="plot.data", :data="tableData", :threshold="threshold", v-model="selected")
+  anova-table.horizontalScrollable(v-if="plot.data", :data="tableData", :threshold="threshold", v-model="selected")
```
waxlamp, line 92: The linter says this line is too long. The fix is to split the attributes out over several lines.

mvandenburgh: Fixed the line that was too long.

subdavis:
@mvandenburgh please rebase from master to resolve the CI errors.

web/src/components/vis/AnovaTableTile.vue
```diff
@@ -89,7 +89,8 @@ vis-tile-large(v-if="plot", title="Anova Table", :loading="plot.loading", expand
     v-btn(flat, dark, block, @click="downloadTable")
       v-icon.mr-2 {{ $vuetify.icons.save }}
       | Download Table
-  anova-table.horizontalScrollable(v-if="plot.data", :data="tableData", :threshold="threshold", v-model="selected")
+  anova-table.horizontalScrollable(v-if="plot.data", :data="tableData",
+      :threshold="threshold", v-model="selected")
```
waxlamp, line 93: When we break a long Pug line, we tend to go all the way: put each attribute on a single line. (But I can't remember if we are supposed to retain the trailing commas in this situation--@subdavis?)

So something like:
```pug
elem(
    v-if="...",
    :data="...",
    .
    .
    .
)
```

subdavis:
If you take a look at the files changed, you can see that it contains 449 additions and 363 deletions: https://github.com/girder/viime/pull/516/files

Those aren't your code changes, so it's a best practice to update your branch so it only includes the changes you made.  You'll want to use `git rebase`  instead of merge, and generally when you rebase you want to squash your own commits first. 

If you checkout your branch, you can do something like this:

```
git checkout master
git pull
git checkout fix-anova-horizontal-scroll
git rebase -i master
# you'll get a list of commits.  Here, we can squash
# leave the first as 'pick' and use `fixup` on the rest.
# so replace `pick` with `f`
# close the editor.  The rebase will fail at 5/8, and you'll have to resolve a conflict in AnovaTableTile.vue
# fix the conflict by hand
git add -u # stage the conflict fix
git rebase --continue # you should be done!
git log # verify that you now have a single commit ahead of master
```

mvandenburgh:
I think it is fixed now, but I had to force push after rebasing. Is there a better way to do that?

subdavis:
> I think it is fixed now, but I had to force push after rebasing. Is there a better way to do that?

No, a force push is always required after a rebase because you're overwriting history rather than appending new commits.  If you're afraid of losing your work, you can always checkout a copy branch before the rebase so you have your changes saved locally.

danlamanna:
I would also strongly recommend using --force-with-lease over the regular
--force.

waxlamp:
> > I think it is fixed now, but I had to force push after rebasing. Is there a better way to do that?
> 
> No, a force push is always required after a rebase because you're overwriting history rather than appending new commits. If you're afraid of losing your work, you can always checkout a copy branch before the rebase so you have your changes saved locally.

Just for information's sake: there is also the "reflog" to help resurrect seemingly lost commits, but that's sort of the last guardrail before you plummet off the mountain. I myself, for instance, have never had to dive into the reflog :crossed_fingers:

waxlamp:
> I would also strongly recommend using --force-with-lease over the regular --force.
> […](#)
> On Thu, May 28, 2020, 1:18 PM Brandon Davis ***@***.***> wrote: I think it is fixed now, but I had to force push after rebasing. Is there a better way to do that? No, a force push is always required after a rebase because you're overwriting history rather than appending new commits. If you're afraid of losing your work, you can always checkout a copy branch before the rebase so you have your changes saved locally.

Fantastic, TIL. Thanks Dan!

PR #517, opened 2020-05-27
Upgrade python dependencies and fix new linting errors

jbeezley:


jbeezley:
It was just locked to older versions.  In particular, matplotlib 3.1.1, which didn't have binary wheels for python 3.8.

PR #518, opened 2020-05-27
Fix narrow volcano plot "Group Combination" panel

mvandenburgh:
I set a width of 100px for each value (the panel has a width of 200px), so now they are truncated by ellipsis and there's a tooltip on each one that shows the full value when hovered over.

### DIFF
--- a/web/src/components/toolbar/ToolbarOption.vue
+++ b/web/src/components/toolbar/ToolbarOption.vue
@@ -41,13 +41,25 @@ div
         v-radio.wide.mr-0(v-for="m in options",
             :value="m.value", :key="m.value")
           template(#label)
-            span.grow {{ m.label }}
-            help-dialog(v-if="m.helpText", :title="`${m.label} ${title}`",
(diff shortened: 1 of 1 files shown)


subdavis:
@mvandenburgh please rebase from master to resolve the CI errors.

web/src/components/toolbar/ToolbarOption.vue
```diff
@@ -41,13 +41,26 @@ div
         v-radio.wide.mr-0(v-for="m in options",
             :value="m.value", :key="m.value")
           template(#label)
-            span.grow {{ m.label }}
-            help-dialog(v-if="m.helpText", :title="`${m.label} ${title}`",
+            span.grow.groupCombinationContainer(
+                v-bind:title="m.label",
+                style="overflow: hidden") {{ m.label }}
```
dchiquit, line 46:
```suggestion
            span.grow.groupCombinationContainer(v-bind:title="m.label")
              | {{ m.label }}
```
the `style="overflow: hidden"` is redundant with the class.

PR #524, opened 2020-05-28
Add search to correlation network

mvandenburgh:
The search function works, but there is an issue where the ForcedDirectedGraph resets itself back to its original position if another node is searched for. 

w45dL8gzS9

### DIFF
--- a/web/src/components/vis/CorrelationTile.vue
+++ b/web/src/components/vis/CorrelationTile.vue
@@ -38,7 +38,14 @@ export default {
       search: [], // nodes being searched for
       currentUserInput: '', // the text currently typed in search box
       searchBarResults: new Set(), // current results from autocomplete search box
-      excludedSearchBarResults: [],
+      excludedSearchBarResults: new Set(),
+      highlightedItems: new Set(),
+      depthOptions: [
+        { value: 0, text: 'Show all' },
+        { value: 1, text: 'Show 1' },
+        { value: 2, text: 'Show 2' },
+        { value: Infinity, text: 'Show reachable' },
+      ],
       searchNodeVisibility: 0,
     };
   },
@@ -78,13 +85,6 @@ export default {
             value: Math.abs(d.value),
           }));
     },
-    searchResults() {
-      // if input is empty, don't highlight any nodes
-      if (!this.currentUserInput || this.currentUserInput.trim() === '') {
-        return [];
-      }
-      return Array.from(this.searchBarResults);
-    },
     nodeCount() {
       return this.nodes.length;
     },
@@ -95,10 +95,26 @@ export default {
       return this.searchNodeVisibility;
     },
   },
+
+  watch: {
+    async currentUserInput(newval) {
+      if (newval) {
+        await this.$nextTick(); // wait for the filter function to populate the result set
+        const highlighted = new Set(this.searchBarResults);
+        this.excludedSearchBarResults.forEach(item => highlighted.delete(item));
+        this.highlightedItems = highlighted;
+      } else {
+        this.highlightedItems = new Set();
+      }
+    },
+  },
+
   methods: {
     searchFilter(item, queryText, itemText) {
-      const match = itemText.toLocaleLowerCase().indexOf(queryText.trim().toLocaleLowerCase()) > -1
-                    && !this.excludedSearchBarResults.includes(itemText);
+      const match = (itemText
+        .toLocaleLowerCase()
+        .indexOf(queryText.trim().toLocaleLowerCase()) > -1
+      ) && !this.excludedSearchBarResults.has(itemText);
       if (match) {
         this.searchBarResults.add(itemText);
       } else {
@@ -106,25 +122,19 @@ export default {
       }
       return match;
     },
-    clearSearch(event) {
-      event.preventDefault();
+    clearSearch() {
       this.search = [];
       this.currentUserInput = ''; // clear search box
       this.searchBarResults = new Set(); // unhighlight search result nodes
     },
+    unselect(val) {
+      this.search.splice(this.search.indexOf(val), 1);
+    },
     clearExcludedNodes() {
-      this.excludedSearchBarResults = [];
+      this.excludedSearchBarResults = new Set();
     },
     removeNodeFromSearchResults(event, data) {
-      event.stopPropagation();
-      this.excludedSearchBarResults.push(data.item);
-      // add char to search box to trigger rerender
-      // of search results.
-      // TODO: find better way to do this
-      if (!this.currentUserInput) {
-        this.currentUserInput = '';
-      }
-      this.currentUserInput = this.currentUserInput.concat('\n');
+      this.excludedSearchBarResults = new Set([...this.excludedSearchBarResults, data.item]);
     },
   },
 };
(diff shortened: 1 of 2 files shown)


jeffbaumes:
It seems maybe there needs to be a rebase of this PR? Unrelated commits from sgratzl and jbeezley are in here.

jeffbaumes:
This is looking good. A few comments:

* The overall UI design for this looks great, and seems to fit right into the sidebar style of components around it.
* It is implemented as an auto-select. This is very helpful to guide the user, but it's unclear whether an arbitrary string can be searched. It may be helpful to highlight several metabolites that match a particular substring. Counterpoint: What if a metabolite name is a strict subset of another? Then there would be no way of selecting the shorter name without also selecting the longer matching one. Maybe selecting from the list can "lock in" to a single metabolite but typing free-form does not?
* Related to this is the possibility of highlighting sets of nodes as you type, which would make things more dynamic. I believe our sets of metabolites are normally small enough (few thousand) to do this without performance concerns. We could debounce, say every 500ms. Obviously this would make the re-layout issue even worse so we'd need to get to the bottom of that.

mvandenburgh:
The reset layout issue has been fixed; I had to make some changes to the way the ForceDirectedGraph component works to prevent the reset from happening. I'll start implementing on the suggestions @jeffbaumes made.

jeffbaumes:
Sounds like a good plan @mvandenburgh. I also notice that you've rebased, so now  the scope of the change is clear (~30 lines right now). The code change is looking clean enough to me so far, so I think you are on the right track. @subdavis I assume you will do final review and sign off once this PR is fully ready.

mvandenburgh:
@subdavis I think this should be all set for you to review, I implemented all the suggestions by @jeffbaumes, namely "selecting from the list can "lock in" to a single metabolite but typing free-form does not" and highlighting the relevant nodes as you type in the search box.

jeffbaumes:
This draft status is new to me, but I believe pressing the "Ready for review" button will add an extra signal that this is no longer WIP, so @mvandenburgh you can go ahead and do that.

web/src/components/vis/CorrelationTile.vue
```diff
@@ -35,6 +35,10 @@ export default {
       correlation_methods,
       metaboliteFilter: null,
       metaboliteColor: null,
+      search: null, // nodes being searched for
+      currentUserInput: null, // the text currently typed in search box
+      searchBarResults: [], // current results from autocomplete search box
+      nodeCount: 0, // total number of nodes in graph
```
subdavis, line 41: seems like this would be good to refactor as a computed property.

It's perfectly valid to have computed properties that are derived from other computed properties.

subdavis: Also, these are helpful comments.  Documenting these really helps me confirm that the logic matches the intent.

web/src/components/vis/CorrelationTile.vue
```diff
@@ -73,6 +77,32 @@ export default {
...
     },
+    searchResults() {
+      if (!this.currentUserInput || this.currentUserInput.length === 0) {
+        return [];
+      }
+      return this.searchBarResults;
+    },
+  },
+  mounted() {
+    this.nodeCount = this.nodes.length;
```
subdavis, line 88: See above.

web/src/components/vis/CorrelationTile.vue
```diff
@@ -73,6 +77,32 @@ export default {
             value: Math.abs(d.value),
           }));
     },
+    searchResults() {
+      if (!this.currentUserInput || this.currentUserInput.length === 0) {
+        return [];
+      }
+      return this.searchBarResults;
```
subdavis, line 84: I would probably make `currentUserInput` default to `''`, so you don't have to think about more than 1 data type.  It's my understanding that `null` and `''` carry the same meaning here: search box empty

Then, you can either check truthiness or you can use strict equality `if (this.currentUserInput === '') {...`

I think strict equality is nicer personally.

web/src/components/vis/CorrelationTile.vue
```diff
@@ -73,6 +77,32 @@ export default {
...
+        }
+        return true;
+      }
+      if (this.searchBarResults.includes(itemText)) {
+        this.searchBarResults.splice(this.searchBarResults.indexOf(itemText), 1, '');
+        if (this.searchBarResults.length > this.nodeCount) {
+          this.searchBarResults = this.searchBarResults.filter(result => result !== '');
+        }
+      }
+      return false;
```
subdavis, line 104: You've gone to some trouble to implement `Set()` here. I think this function would read much more cleanly if you just make `searchBarResults` an instance of `Set`.

That, and the added bonus that set operations are `O(1)` rather than the `O(N)` you get from `includes` and `splice`.

subdavis: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

web/src/components/vis/CorrelationTile.vue
```diff
@@ -111,7 +151,8 @@ vis-tile-large.correlation(v-if="plot", title="Correlation Network", :loading="p
     force-directed-graph(:edges="edges", :nodes="nodes",
         :link-distance="linkDistanceAsNumber", :show-node-labels="showNodeLabels",
         :show-edge-labels="showEdgeLabels",
-        :min-stroke-value="min_correlation")
+        :min-stroke-value="min_correlation",
+        :search="search", :filtered-items="searchResults")
```
subdavis, line 155: newline between props.

Pug is annoying...

web/src/components/vis/ForceDirectedGraph.vue
```diff
@@ -46,6 +46,16 @@ export default {
       default: 0,
       required: false,
     },
+    search: {
+      type: Array,
+      default: null,
+      required: false,
```
subdavis, line 52: Either `default` or `required` alone will do.  It's redundant to provide both.

web/src/components/vis/ForceDirectedGraph.vue
```diff
@@ -76,9 +79,20 @@ export default {
         .clamp(true);
     },
   },
+  watch: {
+    search: function highlightSearchedNode(searchedNodes) {
```
subdavis, line 83: Prefer ES6 property value shorthand in vue functions.

I appreciate that you documented the function's purpose with its name.  Perhaps just move that into a comment inside the method?

```
watch: {
   search() {
     ...
  },
  ...
}
```

web/src/components/vis/CorrelationTile.vue
subdavis: To clarify, Vue doesn't have reactivity for the `Map`and `Set` data types.  You'll still have to copy values over from the set into an array, so maybe a new `searchBarResultsSet` for tracking changes, then `this.searchBarResults = Array.from(this.searchBarResultsSet)`;

subdavis, line 104:
```suggestion
      const match = itemText.toLocaleLowerCase().indexOf(queryText.toLocaleLowerCase()) > -1;
      if (match) this.searchBarResultsSet.add(itemText);
      else this.searchBarResultsSet.remove(itemText);
      this.searchBarResults = Array.from(this.searchBarResultsSet)
      return match;
```

Naturally my formatting is terrible, but that's the idea.

jeffbaumes:
Ideas from the peanut gallery today to implement before merging:

* checkbox alignment
* clear all button
* closable pills

web/src/components/vis/CorrelationTile.vue
```diff
@@ -89,13 +90,27 @@ export default {
...
   methods: {
     searchFilter(item, queryText, itemText) {
-      const match = itemText.toLocaleLowerCase().indexOf(queryText.toLocaleLowerCase()) > -1;
-      if (match) this.searchBarResults.add(itemText);
-      else this.searchBarResults.delete(itemText);
+      const match = itemText.toLocaleLowerCase().indexOf(queryText.toLocaleLowerCase()) > -1
+                    && !this.excludedSearchBarResults.has(itemText);
+      if (match) {
+        this.searchBarResults.add(itemText);
+      } else this.searchBarResults.delete(itemText);
```
subdavis, line 97: Probably should wrap this in brackets.  I know you just followed my bad formatting, sorry.

web/src/components/vis/CorrelationTile.vue
```diff
@@ -132,8 +147,14 @@ vis-tile-large.correlation(v-if="plot", title="Correlation Network", :loading="p
...
             @change="clearSearch")
+          template(v-slot:item='data')
+            v-icon.closePillButton(v-text="'mdi-alpha-x-circle'",
+                style="padding-right: 12px;",
+                @click="(e) => removeNodeFromSearchResults(e, data)")
+            v-list-item-content(v-text="data.item")
+
       span.searchBarContainers
-        v-icon(style="color: red;" @click='search = []', v-text="'mdi-cancel'")
+        v-icon(@click='search = []; clearSearch();', v-text="'mdi-delete'")
```
subdavis, line 157: Move `search = []` into the clearSearch method.

mvandenburgh:
In addition to fixing the bugs I also made it so the dropdown menu isn't ever disabled, since I think it is reasonable for the user to configure that before searching for any nodes.

PR #525, opened 2020-06-01
Move client to netlify

mvandenburgh:
The client application has been moved to netlify here - 
https://lucid-galileo-559611.netlify.app/

The location of the server is now specified with an environment variable and the client uses the full URL for its API calls instead of relying on nginx. Requests to the server are still failing, so I think some configuration needs to be done server side like enabling CORS.

### DIFF
--- a/viime/app.py
+++ b/viime/app.py
@@ -2,6 +2,7 @@ import os
 
 import dotenv
 from flask import current_app, Flask, jsonify
+from flask_cors import CORS
 from flask_migrate import Migrate
 from marshmallow import ValidationError
 from webargs.flaskparser import parser
@@ -50,6 +51,9 @@ def create_app(config=None):
 
     config = config or {}
     app = Flask(__name__)
+
+    # enable CORS for all API calls
+    CORS(app, resources={'/api/*': {'origins': '*'}})
 
     app.wsgi_app = ProxyFix(app.wsgi_app)
     app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
--- a/web/src/common/api.service.js
+++ b/web/src/common/api.service.js
@@ -5,7 +5,7 @@ import VueAxios from 'vue-axios';
 const ApiService = {
   init() {
     Vue.use(VueAxios, axios);
-    Vue.axios.defaults.baseURL = '/api/v1';
+    Vue.axios.defaults.baseURL = `${process.env.VUE_APP_SERVER_ADDRESS}/api/v1`;
   },
 
   buildUrl(path, args) {
--- a/web/src/common/api.service.js
+++ b/web/src/common/api.service.js
@@ -3,9 +3,9 @@ import axios from 'axios';
 import VueAxios from 'vue-axios';
 
 const ApiService = {
-  init() {
+  init(serverURL) {
     Vue.use(VueAxios, axios);
-    Vue.axios.defaults.baseURL = `${process.env.VUE_APP_SERVER_ADDRESS}/api/v1`;
+    Vue.axios.defaults.baseURL = serverURL;
   },
 
   buildUrl(path, args) {
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "701eca715a8ed756d4ba4c863e64c4d5b18f739d34ea8fcf68e08beff08f44c6"
+            "sha256": "e903293ee003e5346835e63803c8de87e7fb6538bced10ca93d263123157f3d0"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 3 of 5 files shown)


mvandenburgh:
Oops, the initial PR included commits from another branch as well. Should be fixed now.

viime/app.py
```diff
@@ -59,6 +60,8 @@ def create_app(config=None):
     app.config['MAX_CONTENT_LENGTH'] = int(os.getenv('MAX_FILE_UPLOAD_SIZE', 5 * 1024 * 1024))
     app.config['OPENCPU_API_ROOT'] = os.getenv('OPENCPU_API_ROOT')
 
+    CORS(app)
```
dchiquit, line 63:
```suggestion
    CORS(app, origins="*")
```
The unconfigured Flask CORS plugin will actually block all cross origin requests by default. You need to specify somewhere that all origins are acceptable.

@jbeezley should the allowed origins go in the `.gunicorn_env` instead?

web/.env.production
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+VUE_APP_SERVER_ADDRESS=https://34.227.42.195
```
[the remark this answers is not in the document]
jbeezley: I think what you want to do rather than hard code the domain in the app is to write a [netlify proxy rule](https://docs.netlify.com/routing/redirects/rewrites-proxies/#proxy-to-another-service) to proxy requests to that address.  That will also avoid the need to set up CORS.

mvandenburgh:
Looks like everything is working now, this branch is deployed at https://5ed7c9abb1455f00071b6147--lucid-galileo-559611.netlify.app. @jbeezley there's two ways to configure the proxy rule (see [here](https://docs.netlify.com/routing/redirects/)), is there any preference for using netlify.toml vs _headers and _redirects files? (I did the latter here)

jbeezley:
I've never set up netlify before, so I don't know exactly how it works, but I think this will still serve the static content in the deployed version from our nginx server on EC2.  I *think* what we need to do is change the DNS reference to the netlify server.  This would mean the redirect target needs to be an IP address (or some other domain like `api.viime.org`).  I would just use the IP address for now since we have an elastic ip assigned.

Our sysadmin controls the DNS, so we'll need to contact them to make the change over to the netlify deployment.  Once that is done we can remove the web client build from EC2.

web/public/_headers
```diff
@@ -0,0 +1,2 @@
+/api/*
+  Access-Control-Allow-Origin: "*"
```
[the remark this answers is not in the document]
jbeezley: Yeah, I guess it is necessary.

web/public/_headers
```diff
@@ -0,0 +1 @@
+/api/*  http://34.227.42.195/api/:splat  200
```
jbeezley, line 1: I think you want `Access-Control-Allow-Origin: "*"`.

mvandenburgh: Oops looks like I copied the _redirects file into it, fixed now.

mvandenburgh:
CORS is set up on the server. Some dependencies were updated when I installed `flask-cors`, is that ok?

waxlamp:
> CORS is set up on the server. Some dependencies were updated when I installed `flask-cors`, is that ok?

That is not generally a problem. The changes stemming from inclusion of `flask-cors` are minimal and necessary.

(For what it's worth, we are using `flask-cors` on Multinet, another project with a Flask-based backend component.)

mvandenburgh:
@jbeezley all set. I set the environment variable in the Netlify settings to that URL.

jbeezley:
Perfect, we'll need to wait on the DNS changes, but I think this can go in as is.  Would you mind squashing the commits?

mvandenburgh:
> Perfect, we'll need to wait on the DNS changes, but I think this can go in as is. Would you mind squashing the commits?

Done.

jbeezley:
@dchiquit Could you resolve your review?

web/src/common/api.service.js
```diff
@@ -5,7 +5,7 @@ import VueAxios from 'vue-axios';
 const ApiService = {
   init() {
     Vue.use(VueAxios, axios);
-    Vue.axios.defaults.baseURL = '/api/v1';
+    Vue.axios.defaults.baseURL = `${process.env.VUE_APP_SERVER_ADDRESS}/api/v1`;
```
subdavis, line 8: Maybe set this in `main.js` and pass it through the init function so that it's easier to find?  Someone looking for the API base url config would start there.

Could you also use `(new URL('/api/v1', process.env.VUE_APP_SERVER_ADDRESS)).href` ?

It handles issues where a trailing slash conflicts with a slash prefix, and throws an exception if the url is invalid.  https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

It doesn't accept an empty string or slash to indicate same origin, but I don't think that case applies anymore anyway.  if needed, `window.location.href` can be input as the base url (and it will be trimmed appropriately).

jbeezley:
api.viime.org is now working (e.g. see https://api.viime.org/api/v1/status)

PR #526, opened 2020-06-02
Fix BoxPlot render error

dchiquit:
BoxPlots are rendered twice consectively, once when mounted and again
after the data is loaded. There is a 200ms D3 transition that is applied
during render. When the transition from the first render has not
completed before the second render starts, it is interrupted. D3 throws
a promise exception whenever a transition is interrupted. This can
safely be ignored.

There is also some spaghetti removal around triggering updates in
BoxPlots.

### DIFF
--- a/web/src/components/vis/BoxPlot.vue
+++ b/web/src/components/vis/BoxPlot.vue
@@ -76,10 +76,16 @@ export default {
       return !this.groups ? this.scaleY.bandwidth() : this.scaleGroup.bandwidth();
     },
   },
-
+  watch: {
+    rows() {
+      this.update();
+    },
+  },
+  mounted() {
+    this.update();
+  },
   methods: {
     update() {
-      //
       // Compute the total variance in all the PCs.
       const svg = select(this.$refs.svg);
       this.axisPlot(svg);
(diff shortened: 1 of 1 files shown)


web/src/components/vis/BoxPlot.vue
```diff
@@ -122,63 +128,57 @@ export default {
...
-            ? `<title>${d.name}: ${f(d.whiskers[0].start)} (q1-iqr*1.5) - ${f(d.fiveNums[1])} (q1) = ${count(d.values, d.whiskers[0].start, d.fiveNums[1])} Items</title>`
-            : `<title>${d.name}: ${f(d.fiveNums[3])} (q3) - ${f(d.whiskers[1].start)} (q3+iqr*1.5) = ${count(d.values, d.fiveNums[3], d.whiskers[1].start)} Items</title>`));
-      }).catch(() => {
-        throw new Error('BoxPlot draw failed');
-      });
+        .call(layout)
+        .end()
+        // silently catch any transition interruptions from consecutive updates
+        .catch(() => {})
+        .then(() => {
```
dchiquit, line 138: From here down is completely unchanged. I removed the `finished` variable which changed the indentation for this whole block.

web/src/components/vis/BoxPlot.vue
```diff
@@ -122,63 +128,57 @@ export default {
...
     },
   },
 };
 </script>
 
 <template lang="pug">
 .main(v-resize:throttle="onResize")
-  svg(ref="svg", :width="width", :height="height", xmlns="http://www.w3.org/2000/svg",
-      :data-update="reactiveUpdate")
+  svg(ref="svg", :width="width", :height="height", xmlns="http://www.w3.org/2000/svg")
```
dchiquit, line 181: `reactiveUpdate` is a computed property which is solely used to trigger re-renders. It's not at all necessary, so I'm removing it here. I will also be removing it from everywhere else in an upcoming review.

PR #527, opened 2020-06-02
Fix race condition NPE

dchiquit:
There is a race condition when loading the initial dataset using the
`LOAD_DATASET` action. Components will use the `loadDataset` mixin and
assume that the `dataset` in the store is non-null. `loadDataset`
triggers the `LOAD_DATASET` action, which asynchronously loads the
dataset. Normally this works fine, but during abnormally long loads or
unlucky scheduling `dataset` will be accessed before it is loaded,
resulting in an NPE in a random part of the app.

We already have an `INITIALIZE_DATASET` mutation that is used internally by `LOAD_DATASET`, so the solution is to
simply have the `loadDataset` mixin invoke it before triggering the
`LOAD_DATASET` action. Mutations are synchronous, so `dataset` is now
guaranteed to be initialized.

### DIFF
--- a/web/src/store/mutations.type.js
+++ b/web/src/store/mutations.type.js
@@ -1,3 +1,4 @@
+export const INTIALIZE_DATASET = 'initialize_dataset';
 export const REMOVE_DATASET = 'remove_dataset';
 export const SET_PLOT = 'set_plot';
 export const SET_SELECTION = 'set_selection';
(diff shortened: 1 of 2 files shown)


waxlamp:
Is the idea that invoking the initialize dataset mutation averts the NPE (thus solving the underlying problem), and when the subsequent load dataset action completes, then data flows to the visualizations?

dchiquit:
Correct. `INITIALIZE_DATASET` puts in a valid dataset object that has no data in it.

PR #528, opened 2020-06-02
Clean up axisPlot reactivity

dchiquit:
Components which used `axisPlot` were relying on an awkward misuse of
computed properties to trigger a render immediately after rendering the
first time. I think the original motivation was that the graph SVG
element was not mounted until after the first render. However this is
trivially solved by calling the `update` function which updates the SVG
in a `mounted` hook.

- Remove the `reactiveUpdate` component from all plots which use it
- Fix a bug where putting "" into an integer field breaks the PCA plot
- Fix a path typo (`wilcoxon_volacno`)

### DIFF
--- a/web/src/components/vis/PcaPage/PcaPage.vue
+++ b/web/src/components/vis/PcaPage/PcaPage.vue
@@ -26,6 +26,9 @@ export default {
       pcXval: '1',
       pcYval: '2',
       numComponentsVal: '10',
+      pcX: 1,
+      pcY: 2,
+      numComponents: 10,
       showEllipses: true,
       showCrosshairs: true,
       showCutoffs: true,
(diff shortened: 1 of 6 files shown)


web/src/components/vis/PcaPage/PcaPage.vue
```diff
@@ -36,25 +39,45 @@ export default {
...
+    pcXval: {
+      handler(val) {
+        const pcX = Number.parseInt(val, 10);
+        // eslint-disable-next-line no-restricted-globals
+        if (!isNaN(pcX)) {
+          this.pcX = pcX;
+        }
+      },
+      immediate: true,
+    },
```
dchiquit, line 59: This pattern feels bad, but I couldn't think of anything better for ensuring that when the input is set to the empty string, the value does not change.

dchiquit:
I pulled the `PcaPlot` input validation changes out of here and moved them to https://github.com/girder/viime/pull/531.

PR #531, opened 2020-06-03
Validate pca inputs

dchiquit:
Currently if you remove all the text from one of the PcaPlot number
inputs, the components will fail bloodily due to the empty string not
parsing to a number.

Use watchers to listen for changes to the inputs, and only update the
associated variables when the numbers are valid.

@waxlamp:
> I am not sure I like the fix to protect the system against bad input in the comboboxes though. The use of watchers to conditionally preserve a good value feels like a maintenance burden. I think a better solution would be to keep the computed props as they were before, and fix the visualizations themselves. I assume it would be as simple as not performing the render action if any of the input values are null (and an improvement to the computed props would be to make it explicit that the type of those props is number | null).
> 
> Can we split out that change into a separate commit or a separate branch so we can have a fuller discussion of it?

The most intuitive behavior to me is to have the graphs not change when an invalid input is provided. I don't see a way in Vue to get this behavior without tracking the last "good" state independently, which requires the `watch` syntax that I've already used here. That code could be moved into the plots, but it would have to be added to every plot separately, which would be a lot more maintenance.

If you want to use computed props, there's no way to maintain the original state if the input isn't valid. Computed props are purely functional and don't have access to the former state of the prop.


dchiquit:
I messed up some rebasing, everything should look right now.

web/src/components/vis/PcaPage/PcaPage.vue
```diff
@@ -36,25 +39,45 @@ export default {
...
+          this.pcX = pcX;
+        }
+      },
+      immediate: true,
+    },
+    pcYval: {
+      handler(val) {
+        const pcY = Number.parseInt(val, 10);
+        // eslint-disable-next-line no-restricted-globals
+        if (!isNaN(pcY)) {
```
subdavis, line 64: `Number.isNaN()`  should fix the lint error?

subdavis: https://stackoverflow.com/questions/46677774/eslint-unexpected-use-of-isnan

The behavior varies slightly from `isNaN`, but since the inputs are only `number` or `NaN` guaranteed by parseInt, you should be safe.

waxlamp:
> Currently if you remove all the text from one of the PcaPlot number
> inputs, the components will fail bloodily due to the empty string not
> parsing to a number.
> 
> Use watchers to listen for changes to the inputs, and only update the
> associated variables when the numbers are valid.
> 
> @waxlamp:
> 
> > I am not sure I like the fix to protect the system against bad input in the comboboxes though. The use of watchers to conditionally preserve a good value feels like a maintenance burden. I think a better solution would be to keep the computed props as they were before, and fix the visualizations themselves. I assume it would be as simple as not performing the render action if any of the input values are null (and an improvement to the computed props would be to make it explicit that the type of those props is number | null).
> > Can we split out that change into a separate commit or a separate branch so we can have a fuller discussion of it?
> 
> The most intuitive behavior to me is to have the graphs not change when an invalid input is provided. I don't see a way in Vue to get this behavior without tracking the last "good" state independently, which requires the `watch` syntax that I've already used here. That code could be moved into the plots, but it would have to be added to every plot separately, which would be a lot more maintenance.
> 
> If you want to use computed props, there's no way to maintain the original state if the input isn't valid. Computed props are purely functional and don't have access to the former state of the prop.

My proposal was to leave the computed props as they were (still purely functional, and reporting `null` or `undefined` in case of bad input values) and defer the work of preserving the visualization state to the vis components themselves.

I bounced this off @subdavis and he says it feels like a leaky abstraction to have the vis components have to react to missing input data, and he's probably right. So I'm ok with the approach here (even though it still feels weird). Thanks for tackling this, @dchiquit.

web/src/components/vis/PcaPage/PcaPage.vue
waxlamp: What about `window.isNaN()`?

web/src/components/vis/PcaPage/PcaPage.vue
```diff
@@ -36,25 +39,43 @@ export default {
...
+          this.pcY = pcY;
+        }
+      },
+      immediate: true,
+    },
+    numComponentsVal: {
+      handler(val) {
+        const numComponents = Number.parseInt(val, 10);
+        // eslint-disable-next-line no-restricted-globals
+        if (!isNaN(numComponents)) {
```
subdavis, line 72: @dchiquit u missed 1 :P

PR #537, opened 2020-06-09
suggestions for search controls

subdavis:
I'm not done yet, but here's sort of where I'm headed.

I think the coloring of the hide button still needs to change, and the icon itself should probably be the `mdi-eye-off` instead of an X.  

## Before
Screenshot from 2020-06-09 09-24-44

## After
Screenshot from 2020-06-09 09-23-27

### DIFF
--- a/web/src/components/vis/CorrelationTile.vue
+++ b/web/src/components/vis/CorrelationTile.vue
@@ -175,11 +175,10 @@ vis-tile-large.correlation(v-if="plot", title="Correlation Network", :loading="p
             hide-details,
             :filter="searchFilter")
           template(v-slot:item="data")
-            v-chip(
-                close,
-                small,
-                @input="removeNodeFromSearchResults($event, data)")
-              span(v-text="data.item")
+            v-chip(smal)
+              v-icon.closePillButton.pr-2(@click.stop="removeNodeFromSearchResults($event, data)")
+                | mdi-eye-off
+              span {{ data.item }}
       v-btn.my-0.mx-0.mt-2(small, flat, @click="clearSearch")
         v-icon.pr-2 mdi-close
         | Clear Selections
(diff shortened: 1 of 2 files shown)


subdavis:
@mvandenburgh I think this addresses most of the concerns I have

* I've resolved the hacking of the input field with newlines.  See the use of `$nextTick` to wait until the filter function is done running
* I've replaced some of the arrays with sets.  You were using `includes` a lot, which can get expensive.  `has` will be faster.
* Changed the radio buttons to a select to conserve vertical space.
* Your point about the overflow is still valid, but I don't really like the `x` getting hidden because a vertical scroll appears. 

Perhaps we should make this whole sidebar wider, since it's starting to get pretty cramped.

subdavis:
There are still a few issues:

* I think you were correct to put your own icon in with the red highlighting.  I should revert that.
* It looks like there's a problem where you select depth 1, then open the filter controls again and hide or select more metabolites.   When you hide, the whole graph appears again.  When you add new metabolites to the filter, they don't trigger BFS to recompute visibility.

mvandenburgh:
> * I think you were correct to put your own icon in with the red highlighting. I should revert that.
Should we go with a scrollbar approach now for the search box overflow problem, since the user will always be able to see the icon now? 
> * It looks like there's a problem where you select depth 1 (or, anything other than 'all'), then open the filter controls again and hide or select more metabolites.   When you hide, the whole graph appears again.  When you add new metabolites to the filter, they don't trigger BFS to recompute visibility.
I'll look into this, I think I know what is causing it.

I agree with all the other changes you made, the UI looks nicer now. I agree that the sidebar should be widened if possible, whether we do the scrollbar or not.

subdavis:
> Should we go with a scrollbar approach now for the search box overflow problem, since the user will always be able to see the icon now?

I was talking about the options inside the menu popover.  When they're in a selected state, the X to the right is still probably ideal?

See my latest change.

subdavis:
Actually, I've moved the icon to the right for both options.

PR #538, opened 2020-06-09
Remove custom classes from pretreatment menu

subdavis:
Remove a ton of hacky CSS that will have to be gutted before a vuetify upgrade anyway.

This PR damages UX in favor of maintainability.

Notice how sub-lists aren't nicely indented anymore (like Impute Table).

Screenshot from 2020-06-09 15-02-01

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "e903293ee003e5346835e63803c8de87e7fb6538bced10ca93d263123157f3d0"
+            "sha256": "701eca715a8ed756d4ba4c863e64c4d5b18f739d34ea8fcf68e08beff08f44c6"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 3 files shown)


jeffbaumes:
See #542 for suggested changes.

> That's a vuetify design choice, and I agree with the default behavior. The direction of the arrow tells you what will happen when the arrow is clicked.

Agree. I just had the mental memory of how it worked before and saw it as a bug, but I now see it as a feature to change the arrow directions to normalize to the MD spec.

mvandenburgh:
My issue was addressed and the UI looks better now, looks good to me.

subdavis:
> Any way to get rid of the small space on top of the list?

Done

> hover

I couldn't find a way to do this.  `inactive` prop gives the styling we want, but it also disables click and router-link behaviors.  Because the list group adds the right drop-down icon, I can't even adjust the width to make the left side fit into the hovered background.

jeffbaumes:
> > hover
> 
> I couldn't find a way to do this. `inactive` prop gives the styling we want, but it also disables click and router-link behaviors. Because the list group adds the right drop-down icon, I can't even adjust the width to make the left side fit into the hovered background.

I'm sure you've spent all the time I'd want spent on this for now. Please create an issue for the backlog that we may or may not get to. Thanks for looking into it.

PR #541, opened 2020-06-10
Replace ORM column model with dataclass

jbeezley:
@mvandenburgh This is a start of what I had in mind for the refactor.  I only made a few of the modifications for the column information, but tried to add comments where there are similar changes need to happen elsewhere.

The changes here are designed to provide as close as possible to a compatible interface to the ORM classes.  This should help minimize the number of changes required to getting a working version of the refactored codebase.  For performance concerns, we may ultimately want to make more invasive changes, but that can wait until later.

Feel free to start from this branch, or just use it as a guide.  I can also add in more comments explaining what is going on if that would help.

### DIFF
--- a/viime/imputation.py
+++ b/viime/imputation.py
@@ -2,14 +2,12 @@ from typing import Dict, List, Tuple
 
 import pandas as pd
 
-from viime.cache import region
 from viime.opencpu import opencpu_request
 
 IMPUTE_MNAR_METHODS = ['zero', 'half-minimum']
 IMPUTE_MCAR_METHODS = ['random-forest', 'knn', 'mean', 'median']
 
 
-@region.cache_on_arguments()
 def impute_missing(table: pd.DataFrame, groups: pd.DataFrame,
                    mnar='zero', mcar='random-forest', p_mnar=0.7,
                    p_mcar=0.4) -> Tuple[pd.DataFrame, Dict[str, List[str]]]:
--- a/viime/models.py
+++ b/viime/models.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from dataclasses import dataclass
 from datetime import datetime
 from io import BytesIO
 from pathlib import Path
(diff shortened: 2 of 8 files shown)


mvandenburgh:
@jbeezley I think the initial refactor is done. 
I had to make some temporary changes to things outside the model class as well as a few tests in the last two commits; these were mostly due to the fact that you can't set individual rows/columns anymore with how the row/column setters are set up right now.

mvandenburgh:
It seems that the CI was failing because it uses python 3.6 (which doesn't have dataclasses), so I removed 3.6 from the tox configuration file.

viime/imputation.py
```diff
@@ -2,14 +2,13 @@
...
 
-from viime.cache import region
 from viime.opencpu import opencpu_request
 
 IMPUTE_MNAR_METHODS = ['zero', 'half-minimum']
 IMPUTE_MCAR_METHODS = ['random-forest', 'knn', 'mean', 'median']
 
 
-@region.cache_on_arguments()
+# @region.cache_on_arguments()
```
jbeezley, line 11: Might as well remove these comments.

viime/models.py
```diff
@@ -537,30 +550,14 @@ class ModifyLabelListSchema(Schema):
...
-
-
 # Cached class method implementations:
 # These are placed here because dogpile does not allow you
 # to include the class instance in the cache key.  :'(
-@region.cache_on_arguments()
-def _read_csv_file(path: str, **kwargs):
-    return pandas.read_csv(path, **kwargs)
-
+# TODO: remove rest of cached class methods
```
jbeezley, line 556: Might as well move these functions into the `CSVFile` class.  Also, `_headers` and `_keys` logic can be moved directly into the property methods.

mvandenburgh: The `_headers` and `_keys` functions each take in parameter so they couldn't be moved into the property methods, so I just moved them into the`CSVFile` class as-is. Not sure if that is the ideal solution though.

jbeezley: Okay, that works.

jbeezley:
LGTM, but I can't actually approve it because I opened the PR.  Can someone else hit the button?

PR #560, opened 2020-06-16
Fix filter node bugs

mvandenburgh:
This fixes two bugs on viime.org now - the "Filter Node" dialog box does not work, and "selected" nodes are not highlighted in orange as they should be.

### DIFF
--- a/web/src/components/vis/ForceDirectedGraph.vue
+++ b/web/src/components/vis/ForceDirectedGraph.vue
@@ -105,7 +105,7 @@ export default {
     search(searchedNodes) {
       // highlights nodes being searched for
       const nodes = select(this.$refs.svg).select('g.nodes').selectAll('g');
-      nodes.select('circle').style('fill', (d) => (searchedNodes.includes(d.id) ? 'red' : d.color));
+      nodes.select('circle').style('fill', (node, index) => (searchedNodes.includes(node.id) ? 'red' : this.nodes[index].color));
       this.showNodesWithinPathLength(this.search, this.visibleNodes);
     },
     highlightedItems(highlightedItems) {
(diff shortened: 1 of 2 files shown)


web/src/components/vis/ForceDirectedGraph.vue
```diff
@@ -119,13 +119,26 @@ export default {
...
+      const nodes = select(this.$refs.svg).select('g.nodes').selectAll('g').select('circle');
+      const edges = select(this.$refs.svg).select('g.edges').selectAll('g').select('line');
+      nodes.style('visibility', (node) => (newNodeSet.has(node.id) ? 'visible' : 'hidden'));
+      edges.style('visibility', (edge) => (newNodeSet.has(edge.source.id) && newNodeSet.has(edge.target.id) ? 'visible' : 'hidden'));
+      nodes.style('color', (node) => node.color);
+    },
+  },
+  updated() {
+    const nodes = select(this.$refs.svg).select('g.nodes').selectAll('g').select('circle');
+    nodes.style('fill', (node, index) => (this.search.includes(node.id) ? 'red' : this.nodes[index].color));
```
subdavis, line 141: I'm losing track of what's going on here.  Why is updated needed instead of using a watcher?  Using updated obfuscates the reader's ability to figure out what is intended to cause this side-effect.

subdavis:
> but we were discussing that eventually there should never be a 'red' fill which overrides other colors

Do you mean using a red outline to indicate both selected and temporarily a member of search autofill results?

jeffbaumes:
> > but we were discussing that eventually there should never be a 'red' fill which overrides other colors
> 
> Do you mean using a red outline to indicate both selected and temporarily a member of search autofill results?

That is what I was thinking. Another option is to swap behavior - temporary search fills in red, while "searched" nodes are outlined. That way, colors are only fully obscured temporarily, not in the final view that might be downloaded as PNG to go in a report, etc.

web/src/components/vis/ForceDirectedGraph.vue
mvandenburgh: The nodes need to be recolored after each rerender of the graph to preserve the selected color nodes. You're right, a watcher would be much better, just switched it.

mvandenburgh:
> That is what I was thinking. Another option is to swap behavior - temporary search fills in red, while "searched" nodes are outlined. That way, colors are only fully obscured temporarily, not in the final view that might be downloaded as PNG to go in a report, etc.

I agree with this. A node being "searched" has certain implications vs. a node that only appears in the search results, and I think having both look the same might be a source of confusion.

jeffbaumes:
> > That is what I was thinking. Another option is to swap behavior - temporary search fills in red, while "searched" nodes are outlined. That way, colors are only fully obscured temporarily, not in the final view that might be downloaded as PNG to go in a report, etc.
> 
> I agree with this. A node being "searched" has certain implications vs. a node that only appears in the search results, and I think having both look the same might be a source of confusion.

Sure, let's go this route then.

PR #562, opened 2020-06-17
Search feature additions

mvandenburgh:


mvandenburgh:
- The node visibility by depth option now shows the induced subgraph (includes all edges shared among the searched nodes) instead of just the spanning tree resulting from the BFS.
- Holding down shift while clicking a node now adds it to the search box.
- There is now an option to show or hide the searched nodes using the same rules (for ex., hide all nodes within depth 1 instead of showing them).
- I also added the ability to specify the depth instead of only allowing 1 and 2 as viime currently does, since Tom mentioned last week that it is useful to be able to start at a set of nodes and grow the graph. See below:
3z5Z3o1Nh0
This wasn't explicitly requested so it can be removed if it's not needed, since it does add some additional complexity to the UI

- The show only search results option in #558 can be achieved here by setting depth to zero using the previously mentioned feature; I can add an explicit option in the dropdown to do this if desired or if the variable-depth option isn't needed.

mvandenburgh:
@jeffbaumes I addressed your first two comments. 

The CI was failing due to the backend, which is strange since this branch doesn't touch anything there :confused:. But it seems to be fixed now.

PR #565, opened 2020-06-18
Typescript boxplot

subdavis:
In general I think this is how vue+d3 is ideally used.  Vue manages the elements, and d3 manages their properties and positions. D3 excels at things like layouts (circle packing, dendrographs, etc) but it's very bad as a system for managing node lifecycle.

The code I've written here gives us full control over presentation and full coverage with typescript.  Line-count went up a bit, but I find this much more readable than what was here before. 

* Note that this isn't complete.  Grouping and tooltips are missing.  I want to get some feedback before I finish up.
* `useAxisPlot` is a port of the AxisPlot mixin, it's not a brand-new idea I made up.    useAxisPlot follows the old d3 style still for populating the axes.  I think this is OK.
* This does use composition API.   Mixins and typescript go together rather poorly, and I've fount the composition api plugin to be quite stable, so this is a half-step toward vue 3.

A note to my reviewers:  I'm looking for feedback on the general approach here.  I know this may be heresy, but I think it has potential.


web/src/components/vis/BoxPlot.vue
```diff
@@ -70,129 +83,142 @@ export default {
...
+    const scaleX = computed(() => scaleLinear()
+      .domain(xrange.value)
+      .range([0, dwidth.value]));
+    const scaleY = computed(() => scaleBand()
+      .domain(props.rows.map((d) => d.name))
+      // @ts-ignore d3 typings are bad
+      .range([0, dheight.value], 0.1));
+    // const scaleGroup = computed(() => scaleBand()
+    //   // @ts-ignore d3 typings are bad
+    //   .domain(props.groups || []));
```
dchiquit, line 97: delete?

web/src/components/vis/snippets/BoxplotBox.vue
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export default Vue.extend({
+  props: {
+    data: {
+      type: Object as PropType<BoxData>,
+      required: true,
+    },
+  },
+});
```
dchiquit, line 17: I'm not really up to date on the composition API pros/cons, but would it be better to use it for all new components, or are we only using it where there's a very convincing reason to? This component is trivially small so it's not really important here, I'm just curious about our general philosophy.

web/src/components/vis/BoxPlot.vue
```diff
@@ -1,67 +1,80 @@
...
+      type: Array as PropType<Row[]>,
       default: () => [],
     },
     groups: {
-      type: Array,
-      required: false,
-      default: null,
+      type: Array as PropType<Group[]>,
+      default: () => [],
     },
```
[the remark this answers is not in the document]
subdavis: From my comment above, grouping implementation was missing in this round.  Added in 0f11052

web/src/components/vis/BoxPlot.vue
subdavis: See 0f11052

web/src/components/vis/snippets/BoxplotBox.vue
subdavis: So far, I've been following the guidance that it's only worth using if you need mixin-like behavior a-la `useAxisPlot`, otherwise best to stick to object API.

web/src/components/vis/BoxPlot.vue
```diff
@@ -70,129 +83,142 @@ export default {
...
+      });
+
+      const _scaleX = scaleX.value;
+      const _scaleY = scaleY.value;
+
+      return valueStats.map((data) => ({
+        group: data.group,
+        boxes: data.boxes.map((box) => ({
+          name: box.name,
+          transform: `translate(0, ${_scaleY(box.name) + (boxHeight / 2)})`,
```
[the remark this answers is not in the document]
subdavis: This has been updated, but I'm not sure why this would be the case.  Is the transform line still causing an issue for you?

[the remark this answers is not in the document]
subdavis: Found / fixed.

web/src/components/vis/snippets/BoxplotBox.vue
waxlamp: @subdavis, is that guidance specific to Viime and our cold-storage prep? In general I would think abandoning object API would be a long-term goal.

subdavis: > In general I would think abandoning object API would be a long-term goal.

That guidance is general.  Object API is here to stay.  It reads better and it's less flexible than composition which leads to higher quality code in the average case.  

There's a little bit of fighting right now between them because Vue 2's object API passes props that use a reactivity implementation different than how composition works, but in Vue 3 it doesn't matter which style you use, refs will be refs, so you can switch between them freely.

web/src/components/vis/BoxPlot.vue
```diff
@@ -70,129 +90,181 @@ export default {
         }
       });
       return [min, max];
-    },
```
jeffbaumes, line 73: Could we replace this function with a call to [`d3.extent`](https://github.com/d3/d3-array/blob/master/README.md#extent)?

jeffbaumes: Then again maybe not if you will be adding back in grouping, which complicates this function.

subdavis: I would prefer extent if possible.  I'm trying to combine `extent` with `lodash.flatMap` which I think will read better.  I think that should do what you're asking for.

web/src/components/vis/BoxPlot.vue
```diff
@@ -1,198 +1,258 @@
...
+      dheight,
+      axisPlot,
+    } = useAxisPlot(margin, width, height);
+
+    const xrange = computed(() => {
+      if (props.groups.length) {
+        return extent(flatMapDeep(props.rows, (r) => flatMapDeep(r.groups, (g) => g.values)));
+      }
+      return extent(flatMapDeep(props.rows, (r) => r.values || []));
+    }) as Ref<[number, number]>;
```
subdavis, line 80: @jeffbaumes this is the best I could come up with as a means of flattening in order to use `extent`.

jeffbaumes: Yeah, I had realized the grouping would cause an unavoidable conditional for the added level of flattening. I'm good with this.

PR #567, opened 2020-06-24
Hide labels of hidden nodes and edges

### LINKED ISSUE #555: Hide node/edge labels of hidden elements

mvandenburgh:
Resolves #555.

### DIFF
--- a/web/src/components/vis/ForceDirectedGraph.vue
+++ b/web/src/components/vis/ForceDirectedGraph.vue
@@ -125,11 +125,18 @@ export default {
     },
     excludedItems(excluded) {
       // circles nodes in search results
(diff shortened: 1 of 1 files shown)


mvandenburgh:
> This is unfortunately more complex than I was expecting, probably due to D3 controlling visibility and not Vue. But it functions as expected and I don't doubt it is nearly as simple as possible for now.
> 
> We may want to consider a boxplot-ish revamp here that gives more control to Vue so this type of thing happens using Vue's reactivity more naturally.

Yeah the code is getting more and more complex every time I add a new feature. I think using D3 over Vue to control visibility was an oversight on my part due to a lack of knowledge when I started working on it. The node hiding that occurs with the "Node Filter" option in the left panel does actually use Vue's reactivity to control visibility rather than D3, and based on that, a revamp would eliminate a lot of complex code.

PR #575, opened 2020-07-01
Use TypeScript in LoadingsPlot

dchiquit:
Following the example of BoxPlot, update LoadingsPlot to use Typescript.

There was an attempt to move the rendering from D3 into Vue, but that
was stymied by some strange event handling around mouseover tooltips.
For now LoadingsPlot is still using the same rendering code as always.

There was also a small bug where the X axis was being rendered at the
top of the plot instead of the bottom.

This PR is waiting on https://github.com/girder/viime/pull/565 for merge cleanliness. I think it's ready for review though.

### DIFF
--- a/web/src/components/vis/BoxPlot.vue
+++ b/web/src/components/vis/BoxPlot.vue
@@ -70,7 +70,12 @@ export default defineComponent({
       dwidth,
       dheight,
       axisPlot,
-    } = useAxisPlot({ margin, width, height, topAxis: true });
+    } = useAxisPlot({
+      margin,
+      width,
+      height,
+      topAxis: true,
+    });
 
     const xrange = computed(() => {
       if (props.groups.length) {
--- a/web/src/components/vis/BoxPlot.vue
+++ b/web/src/components/vis/BoxPlot.vue
@@ -70,7 +70,7 @@ export default defineComponent({
       dwidth,
       dheight,
       axisPlot,
-    } = useAxisPlot(margin, width, height);
+    } = useAxisPlot({ margin, width, height, topAxis: true });
 
     const xrange = computed(() => {
       if (props.groups.length) {
@@ -250,7 +250,7 @@ export default defineComponent({
 
 <style scoped>
 .boxplot {
-  transition: .4s ease-in-out;
+  transition: 0.4s ease-in-out;
 }
 .axes >>> text {
   font-size: 14px;
--- a/web/src/components/vis/LoadingsPlot.vue
+++ b/web/src/components/vis/LoadingsPlot.vue
@@ -1,18 +1,42 @@
-<template lang="pug">
-.main(v-resize:throttle="onResize")
-  svg(ref="svg", :width="width", :height="height", xmlns="http://www.w3.org/2000/svg")
-    g.master
-      g.axes
-      g.label.x(style="opacity: 0")
-        text x
-      g.label.y(style="opacity: 0")
-        text y
-      g.crosshairs
-      g.plot
-  .tooltip(ref="tooltip")
+<template>
+  <div
+    ref="mainRef"
+    class="main"
+  >
+    <!-- v-resize:throttle="onResize" -->
+    <svg
+      ref="svgRef"
+      :width="width"
+      :height="height"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <g class="master">
+        <g class="axes" />
+        <g class="label x">
+          <text>x</text>
+        </g>
+        <g class="label y">
+          <text>y</text>
+        </g>
+        <g class="crosshairs" />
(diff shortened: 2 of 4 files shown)


subdavis:
Neat!

> There was an attempt to move the rendering from D3 into Vue, but that
was stymied by some strange event handling around mouseover tooltips.
For now LoadingsPlot is still using the same rendering code as always.

What do you think about how it's currently written?   Happy with the traditional D3 approach?

> There was also a small bug where the X axis was being rendered at the
top of the plot instead of the bottom.

This was intentional.  Because the box plot is fixed-height, you might have to scroll to get to the bottom axis.  I thought it looked better to put the axis on top so that it's visible on page load.

dchiquit:
I think the Vue rendering would be cleaner, but it would take some experimentation to replace all the D3 functionality. The tooltip thing I'm sure has a solution, but I didn't want to waste any more time on it. For the loadings plot, the only other thing that D3 provided was the slick transitions, so everything should be much cleaner if we can avoid using D3 altogether. I'm pretty sure we could even get rid of the `update` function, which is my least favorite part of D3.

Regarding the axis, is that how we want all plots to work? All the PCA plots and the volcano plots will hopefully eventually use that mixin, and will inherit that top axis. Personally I think the axis on the bottom is more legible, but I don't feel very strongly. I just fixed it because I thought it was involuntary.

jeffbaumes:
I can see both points. I can see how in the boxplot the top axis makes sense, and in the other plots with a guaranteed aspect ratio the bottom makes sense. Hate to complicate things, but can the mixin take a parameter to specify top or bottom for x axis?

subdavis:
>  I didn't want to waste any more time on it. For the loadings plot, 

I don't really consider transitions as a feature of d3 since it's just a wrapper around the addition of CSS transitions.

> Regarding the axis, is that how we want all plots to work? 

In my head, it was just a special caveat of BoxPlot.vue because it's the only plot we have that doesn't grow/shrink to fill its container, since each vertical bar has a fixed height.

Any dynamically resizeable plots should have their axis on the bottom.  Not that big of a deal to me either.

waxlamp:
> I think the Vue rendering would be cleaner, but it would take some experimentation to replace all the D3 functionality. The tooltip thing I'm sure has a solution, but I didn't want to waste any more time on it. For the loadings plot, the only other thing that D3 provided was the slick transitions, so everything should be much cleaner if we can avoid using D3 altogether. I'm pretty sure we could even get rid of the `update` function, which is my least favorite part of D3.

I don't think we should take it as an implicit goal to replace all D3 rendering with Vue-based equivalents. There are cases where this would be indicated: I've seen graduate-student-written code where the entire UI was built up using D3's DOM manipulation logic, and that's a nightmare for all but the person who wrote the code. But if D3 is being targeted to its actual expert domain--manipulating only the DOM elements involved in a [unit visualization](https://ieeexplore.ieee.org/document/8233127)--then it very likely brings more to the table than the cost it extracts from a Vue-only aesthetic in the codebase.

For instance, sometimes the slick transitions D3 enables are actually informative (though, I admit, in many cases out in the field, they are not). Even when they don't carry useful information, they can be instrumental in getting the customer/user excited and feeling good about the software. Unscientific as that sounds, it is a practical concern we face as we develop these systems. And more generally, the abstractions provided by D3 are quite powerful when it comes to designing a visualization, especially one with specific needs in a specific instance (e.g., both the PCA score plot, and the loadings plot in this project).

I agree with everyone that we have not figured out the best way to mix a Vue component with D3 rendering logic. Part of the problem is that our group contains very few D3 programmers (on this project, I wager it's just Jeff and myself), which means that the D3 style is alien to the group at large. Insofar as this project requires custom visualization, however, I would argue that the cost of using D3 is outweighed by the visualization-related experessiveness it enables.

In any case, I believe there's still something valuable we could learn using Viime as a testbed: how to successfully use a D3 rendering core within a Vue component with a minimum of disruption to the Vue way of doing things. If we can figure this out, it will make an important plank in the "knowledge platform" I see us building.

To make this actionable: I would like to lead a discussion about the problems we're facing in this domain, what the various solutions are, and what we might want to do going forward. Please let me know if you're interested in joining that discussion and I will set something up.

jeffbaumes:
Thanks for the insight @waxlamp. Let's discuss in more detail at next Thursday's meeting. In the meantime, are you comfortable with the content of this PR, or do we need to discuss first?

waxlamp:
> Thanks for the insight @waxlamp. Let's discuss in more detail at next Thursday's meeting.

:+1: 

> In the meantime, are you comfortable with the content of this PR, or do we need to discuss first?

Upgrading to TypeScript is great. I think the PR is good, especially if @subdavis is happy with it. Thanks for this work, @dchiquit!

PR #577, opened 2020-07-02
Migration for row/column refactor

jbeezley:


jbeezley:
@mvandenburgh Could you take a quick look at this?  I ran it against the production database locally and it seems to work.

mvandenburgh:
Looks good to me. I tested it with a local database from before the refactor and it worked.

PR #579, opened 2020-07-02
Set up docker-compose

mvandenburgh:


subdavis:
@mvandenburgh this docker-compose would be useful mainly for running the backend so that front-end developers only had to launch this and `yarn serve`.  What do you think about supporting https://github.com/Kitware/ldc so that the flask server can also be run in development (debug) mode within docker?

mvandenburgh:
> @mvandenburgh this docker-compose would be useful mainly for running the backend so that front-end developers only had to launch this and `yarn serve`. What do you think about supporting https://github.com/Kitware/ldc so that the flask server can also be run in development (debug) mode within docker?

@subdavis Sounds good, I can change it to support that tool.

PR #580, opened 2020-07-02
574 e2e test

JellyWang-3:
574-issue

I could not get the click button test to work. It seems that I would need to play around with the jest-puppeteer-vuetify helper to get the actual button. Is there a way to click a vBtn with the existing puppeteer built-in? Or is that impossible?

Another thing is I tried to define CLIENT_URL as 'http://localhost:8080' but kept getting " Protocol error (Page.navigate): Invalid parameters url: string value expected" error. My quick fix is just put the string back but it seems like there's a better fix.

574-issue2


e2e-test/README.md
```diff
@@ -0,0 +1,116 @@
...
+  it('should behave like this', async () => {
+    // ... test steps ...
+  });
+  it('does the right thing', async () => {
+    // ... test steps ...
+  });
+});
+```
+
+### expect-puppeteer
```
dchiquit, line 49: Can you remove everything from here down?

e2e-test/README.md
```diff
@@ -0,0 +1,116 @@
...
+yarn install
+```
+
+Puppeteer includes a bundled Chromium executable, but not all the various Chrome dependencies.
+This is not generally an issue in development environments where Chrome is already installed.
+Be aware of this when setting up CI or Docker images though.
+
+## Running Tests
+You will need a running instance of the app, both the `web` and `flask` components.
+Assuming the web app is running at `http://localhost:8080`:
```
dchiquit, line 17: I suspect all references to `localhost:8080` will need to be changed to `localhost:8085`

e2e-test/jest.config.js
```diff
@@ -0,0 +1,19 @@
...
+    // Normally, expect-puppeteer is specified by the jest-puppeteer preset
+    'expect-puppeteer',
+    'jest-puppeteer-vuetify',
+    './jest.setup.js',
+  ],
+
+  transformIgnorePatterns: [
+    // Apply Babel transformation to jest-puppeteer-vuetify
+    'node_modules/(?!jest-puppeteer-vuetify/)',
+  ],
```
dchiquit, line 15: This is an artifact from when `jest-puppeteer-vuetify` was in a subdirectory, I think it can be safely removed now.

e2e-test/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "test",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "lint": "eslint .",
```
dchiquit, line 6: Have you tried running `yarn run lint`? Since this was just added it won't get called automatically in CI, so there's no way to know that this new code is actually linted properly.

e2e-test/README.md
JellyWang-3: Do I keep line 49 or no?

e2e-test/package.json
JellyWang-3: I will run `yarn run lint` and fix lint issues

e2e-test/README.md
dchiquit: ah sorry, I left a comment on the wrong line. I'd keep this section, just remove stuff referring to stuff that was in `jest-vuetify-xpaths`. So line 58 and onward should be removed.

PR #582, opened 2020-07-09
WIP set up CI test

JellyWang-3:
https://app.circleci.com/pipelines/github/girder/viime/1628/workflows/7d954986-c1cb-4a90-8b6c-7a29f12085a8/jobs/4118

I have tried different things and it is not working as I expected. I tried to reuse the openCPU image that was provide and I am not sure what this command exactly does.  
`/bin/sh -c "chmod a+w /usr/local/lib/R/site-library && service cron start && /usr/lib/rstudio-server/bin/rserver && apachectl -DFOREGROUND"`
My thought is that this would create another container that has the openCPU in it and then it would allow the backend to work.

The error messages are about commands not found but I don't know how to solve that issue. It seems to me that I have already installed the packages needed and the commands should been running in the background for `yarn serve` and `flask run`

Selection_064

Then probably because of the previous commands are failing, so the `yarn test ` command also failed.


.circleci/config.yml
```diff
@@ -44,6 +44,51 @@ jobs:
...
+      - checkout
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - yarn-packages-{{ checksum "e2e-test/package.json" }}
+      - save_cache:
+          name: Save Yarn Package Cache
+          key: yarn-packages-{{ checksum "e2e-test/package.json" }}
+          paths:
+            - ~/.cache/yarn
```
dchiquit, line 63: This is cool, but premature. Caching is purely for speeding up a slow build by skipping some steps, but this isn't working at all yet. I'd just completely ignore caching until there's a need for it. Personally I'd consider any end to end testing runtime under 5 minutes as perfectly acceptable.
It looks like this configuration wouldn't actually do any useful caching anyway. You restore the cache as the first step (which is fine), then immediately save the cache again without actually installing anything. If anything, I believe the order would be:
* `restore_cache`
* `run: yarn install` (if the cache was loaded, this is fast, otherwise it takes ~2 minutes)
* `save_cache`

Thanks for bringing this up! I'd never heard of this.

.circleci/config.yml
```diff
@@ -44,6 +44,51 @@ jobs:
...
+      - run:
+          working_directory: ./e2e-test
+          name: Install dependencies
+          command: yarn
+      - run:
+          working_directory: ./e2e-test
+          name: Run lint tests
+          command: yarn run lint
+      - run:
+          working_directory: ./e2e-test
```
dchiquit, line 73:
```suggestion
          working_directory: ./web
```
This will navigate to the `e2e-test` directory, then run `yarn serve`. `serve` isn't defined in `e2e-test` though, it's defined in `web`. You can see this step failing [here](https://app.circleci.com/pipelines/github/girder/viime/1629/workflows/023da26e-f4c2-41fe-9cba-cb26975775ce/jobs/4121/parallel-runs/0/steps/0-106).

.circleci/config.yml
```diff
@@ -44,6 +44,51 @@ jobs:
...
+          name: Serve the website
+          command: yarn serve
+          background: true
+      # flask run
+      - run:
+          working_directory: ./e2e-test
+          name: Start the development server
+          command: flask run
+          background: true
```
dchiquit, line 82: This step is failing because you haven't installed anything python related yet, but are already trying to invoke `flask run`. See [dandiarchive](https://github.com/dandi/dandiarchive/blob/master/.circleci/config.yml#L50-L66) for an example of how to install the app prior to serving it.

.circleci/config.yml
JellyWang-3: Oh I see it now.

JellyWang-3:
@dchiquit I am still stuck on this. I don't know how to solve this.
Selection_065
My approach is setting up the python env for flask to run first, then I would run the front end web. I am not sure about the OpenCPU command part and the part that I commented out for docker. After all the set up, I would run the actual tests. 
https://app.circleci.com/pipelines/github/girder/viime/1658/workflows/4fc3f5c4-1032-4561-a4a9-47683d119a95/jobs/4198

.circleci/config.yml
```diff
@@ -44,6 +44,81 @@ jobs:
...
+      #     command: |
+      #         docker build -t viime .
+      #         docker run -it --rm -p 8004:8004 viime
+      #     background: true
+      # serve the front end
+      - run:
+          working_directory: ./web
+          name: Install dependencies
+          command: yarn
+          background: true
```
dchiquit, line 92: This shouldn't be a background task. Background tasks are only for cases where you are serving something and need it to be running simultaneously with something else. It's important that the dependencies are installed before you move on to the next step.

.circleci/config.yml
```diff
@@ -44,6 +44,75 @@ jobs:
...
+      # docker is broken and needs to be fixed but CI succeeded               systemctl --user enable docker
+      - run:
+          working_directory: ./devops
+          name: Run docker
+          command: |
+              docker ps
+              docker build -t viime .
+              docker run -it --rm -p 8004:8004 viime
+          background: true
```
dchiquit: Looks like you have the image working, this step can be removed now.

.circleci/config.yml
```diff
@@ -44,6 +44,75 @@ jobs:
...
+          command: yarn
+      - run:
+          working_directory: ./web
+          name: Serve the website
+          command: yarn serve
+          background: true
+      - restore_cache:
+          name: Restore Yarn Package Cache
+          keys:
+            - yarn-packages-{{ checksum "e2e-test/package.json" }}
```
dchiquit, line 94: After thinking about this I think caching is definitely out of scope for this PR. The [docs](https://circleci.com/docs/2.0/caching/) should be read thoroughly before adding caching; this snippet copied from the `test_web` job will, I think, create cache collisions between the two jobs. It seems like if we're going to be caching the `e2e-test` build files we might as well cache everything. Additionally, the caches can be shared between the different build steps, i.e. the build from `test_web` can be cached and reloaded here in `test_e2e`.
Doing caching properly would take a substantial amount of work to tune up, so it should be done later if CI latency is ever an issue.

.circleci/config.yml
```diff
@@ -141,6 +209,10 @@ workflows:
     jobs:
       - test_python
       - test_web
+      - test_e2e:
+          requires:
+            - test_python
+            - test_web
       - release:
           requires:
             - test_web
```
dchiquit, line 218: These `requires` statements need to be changed to `test_e2e` instead of `web` and `python`. As is, the release step might run successfully before the `test_e2e` step can finish, which would be very bad.

Same change for `publish`.

.circleci/config.yml
```diff
@@ -44,6 +44,74 @@ jobs:
...
+          command: flask run
+          background: true
+      # docker is broken and needs to be fixed but CI succeeded
+      - run:
+          working_directory: ./devops
+          name: Run docker
+          command: |
+            docker build -t viime .
+            docker run -it --rm -p 8004:8004 viime
+          background: true
```
dchiquit, line 81: I see you have the image figured out, I think this can be safely deleted.

.circleci/config.yml
JellyWang-3: Do you mean keep `- test_e2e` in the jobs block, then add `- test_e2e` to both `release` and `publish` 's `requires` block?

dchiquit:
```   
    jobs:
      - test_python
      - test_web
      - test_e2e:
          requires:
            - test_python
            - test_web
      - release:
          requires:
            - test_e2e
          filters:
            branches:
              only: master
      - publish:
          requires:
            - test_e2e
          filters:
            branches:
              only: master
```
We want a chain of jobs that looks like this:
```
test_web    ---> test_e2e --> release
test_python -/             \- publish
```
It just complicates the graph to make `release` and `publish` require `test_web` and `test_python` when in fact we only need to wait for `test_e2e` to finish.

PR #583, opened 2020-07-10
Add an endpoint for ROC analysis

jbeezley:
Note, the endpoint responds with arrays `thresholds`, `sensitivities`, and `specificities`.  I'm not sure precisely how these should be used in the actual ROC plot, but I think it contains all the information needed.

There are also probably a better name for the `column` argument in the analysis, but I'm not sure what it should be.


devops/viime/R/analyses.R
```diff
@@ -152,3 +152,46 @@ clustered_heatmap <- function(measurements) {
...
+
+    # Save predicted values from random forest
+    pred <- rf.model$votes[,1]
+  } else {
+    stop("Invalid method")
+  }
+
+  #############
+  #ROC analysis
+  roc.a <- pROC::roc(group_mask, pred, plot=FALSE)
```
mvandenburgh, line 191: The web client also needs the area under the curve, which I believe can be calculated with `auc(roc.a)`. (https://rdrr.io/cran/pROC/man/auc.html)

jbeezley: I added it in [`530e90e` (#583)](https://github.com/girder/viime/pull/583/commits/530e90e9dcf0a9108432bb47b1de4dfbc03418a8).  The R server won't let you return heterogeneous types, so it returns the scalar value repeated in a list.

PR #585, opened 2020-07-13
Convert ScreePlot to use TypeScript

### LINKED ISSUE #584: Convert ScreePlot to use TypeScript
Modernize ScreePlot in the style of BoxPlot and LoadingsPlot.

dchiquit:
Fixes #584

### DIFF
--- a/web/src/components/vis/ScreePlot.vue
+++ b/web/src/components/vis/ScreePlot.vue
@@ -1,20 +1,41 @@
-<template lang="pug">
-.main(v-resize:throttle="onResize")
-  svg(ref="svg", :width="width", :height="height", xmlns="http://www.w3.org/2000/svg")
-    g.master
-      g.axes
-      g.label.x
-        text {{ xlabel }}
-      g.label.y
-        text {{ ylabel }}
-      g.plot
-        path.line
-        g.points
-        g.cutoffs
-          line.cutoff.cutoff50
-          line.cutoff.cutoff80
-          line.cutoff.cutoff90
-  .tooltip(ref="tooltip")
+
+<template>
+  <div
+    class="main"
+    ref="mainRef"
+  >
+    <!-- v-resize:throttle="onResize"-->
+    <svg
+      ref="svgRef"
+      :width="width"
+      :height="height"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <g class="master">
+        <g class="axes">
+          <g class="label x">
+            <text />
+          </g>
+          <g class="label y">
+            <text />
+          </g>
+        </g>
+        <g class="plot">
+          <path class="line" />
+          <g class="points" />
+          <g class="cutoffs">
+            <line class="cutoff cutoff50" />
+            <line class="cutoff cutoff80" />
+            <line class="cutoff cutoff90" />
+          </g>
+        </g>
+      </g>
+    </svg>
+    <div
+      class="tooltip"
+      ref="tooltipRef"
+    />
+  </div>
 </template>
 
 <style scoped>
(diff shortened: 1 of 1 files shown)


web/src/components/vis/ScreePlot.vue
```diff
@@ -1,20 +1,40 @@
...
-          line.cutoff.cutoff50
-          line.cutoff.cutoff80
-          line.cutoff.cutoff90
-  .tooltip(ref="tooltip")
+<template>
+  <div
+    ref="mainRef"
+    class="main"
+  >
+    <!-- v-resize:throttle="onResize"-->
```
subdavis, line 6: None of the resize controls work anymore.    I feel like we should just get rid of all those zoom controls and the needlessly complex layout styles that drive them.  flex boxes and no resizing would probably be fine.

Screenshot from 2020-07-20 08-59-46

web/src/components/vis/ScreePlot.vue
```diff
@@ -48,105 +68,93 @@ line.cutoff90 {
...
-        numComponentsInternal,
-        dwidth,
-      } = this;
-
-      const labels = [...Array(numComponentsInternal).keys()].map((d) => d + 1);
-      return scalePoint()
+  setup(props) {
+    const svgRef = ref(document.createElement('svg'));
+    const mainRef = ref(document.createElement('div'));
+    const tooltipRef = ref(document.createElement('div'));
```
subdavis, line 113: I still don't know whether it's right to in initialize these to detached elements in order to prevent the "null" case or if the null case is there to protect me from using these before the mounted() hook (or in the event that one of these elements is inside an un-rendered `v-if`)

I definitely think this defeats the purpose of typescript, since we've effectively lied about that the actual type of the ref, but I don't know how much that matters.  I've done this too in other components, FWIW.

web/src/components/vis/ScreePlot.vue
```diff
@@ -156,68 +164,41 @@ export default {
...
-      }
-
-      const radius = 4;
-
-      if (!this.$refs.svg) {
+    });
+
+    function update() {
+      // Recalculate width/height
+      const bb = mainRef.value.getBoundingClientRect();
```
subdavis, line 172: Here, for example, in order to prove to myself this isn't null, I have to look at all the places that call update.

jeffbaumes:
I'd be willing to lose some of that zoom sizing behavior as long as it is decently responsive and all charts are always "reachable", i.e. none go off page where you can't interact or scroll to them.

web/src/components/vis/ScreePlot.vue
dchiquit: I cribbed this from your work in BoxPlot, so I haven't put much thought into it.
That said, I don't see anything wrong with using this:
```
const svgRef = ref() as Ref<HTMLElement>;
```
You'll still get a runtime exception if you try to use the ref directly in `setup`, but as I understand it will be bound properly when any reactive functions are called after `setup` has completed. As long as you don't try to do any funky initializations using refs and make sure to use the `onMounted` hook to trigger the first update, everything should just work.

Regarding https://github.com/girder/viime/pull/589#discussion_r457376662, I personally thing the `onMounted(watchEffect(() => {...}));` is the least bad pattern. D3 requires the elements to be present before it can do it's thing, so it doesn't seem meaningful to be calling an update function prior to mounting.

subdavis: > You'll still get a runtime exception if you try to use the ref directly in setup

It's just not an accurate description of the type of `svgRef`

```
const svgRef = ref() as Ref<HTMLElement | undefined>;
```

This will be a compile error rather than a runtime error if you don't perform the undefined check. 

> onMounted(watchEffect(() => {...})); is the least bad pattern.

Sure, I don't mind that.

web/src/components/vis/ScreePlot.vue
waxlamp: Agreed.

web/src/components/vis/ScreePlot.vue
waxlamp: @subdavis, do you grok the lifecycle process for setting up `ref`s? Casting the undefined values to to `Ref<HTMLElement>` has a certain simple elegance to it _if and only if_ the runtime behavior is in some way guaranteed to perform a kind of "IO monad from Haskell" magic in the background, then pave that magic over with types.

Far more likely is that this would work sort of, and then lead to issues later when, e.g., the Vue people update the type behavior for refs at some point.

I'll reiterate what I wrote in my review of #589: maybe we should file an issue asking about this and see what the Vuefolken say about it.

web/src/components/vis/ScreePlot.vue
```diff
@@ -48,105 +67,93 @@ line.cutoff90 {
...
+} from '@vue/composition-api';
+import { axisBottom, axisLeft } from 'd3-axis';
+import { scalePoint, scaleLinear } from 'd3-scale';
 import { select, event } from 'd3-selection';
 import { format } from 'd3-format';
 import { line } from 'd3-shape';
 import 'd3-transition';
 
-import { axisPlot } from './mixins/axisPlot';
+import useAxisPlot from './use/useAxisPlot';
```
waxlamp, line 81: Is this `use` directory a recommended pattern?

PR #587, opened 2020-07-13
Initial ROC curve visualization

mvandenburgh:
This is an initial implementation of a ROC curve visualization. Depends on #583 .

All the axios code that handles communicating with the new ROC endpoint is temporary, just to get a quick prototype working for the meeting tomorrow. ~~There's also a bug where column names with spaces cause the request to the endpoint to fail, but column names without spaces work fine.~~ Fixed.

### DIFF
--- a/devops/Dockerfile.base
+++ b/devops/Dockerfile.base
@@ -32,6 +32,7 @@ RUN R -e 'install.packages("devtools"); library(devtools)'
 RUN R -e 'library(devtools); devtools::install_github("klutometis/roxygen")'
 RUN R -e 'BiocManager::install("impute", version = "3.9")'
 RUN R -e 'install.packages("missForest")'
+RUN R -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/mnormt/mnormt_1.5-6.tar.gz", repos=NULL, type="source")'
 
 ADD viime /viime
 RUN R -e 'setwd("/viime"); devtools::update_packages(devtools::dev_package_deps()$package, dependencies=NA, upgrade="never");' && rm -fr /viime
--- a/devops/Dockerfile.base
+++ b/devops/Dockerfile.base
@@ -1,4 +1,4 @@
-FROM opencpu/rstudio
+FROM opencpu/rstudio:2.1.7
 
 RUN apt-get update && apt-get install -y \
     libcurl4-openssl-dev \
@@ -13,26 +13,27 @@ RUN dpkg -r --force-all libglib2.0-0 && \
 
 RUN apt-get install libcairo2-dev libxt-dev -y
 
-RUN R -e 'install.packages("XML");' && \
-    R -e 'install.packages("plotly");' && \
-    R -e 'install.packages("BiocManager");' && \
-    R -e 'BiocManager::install("genefilter", version = "3.9")' && \
-    R -e 'BiocManager::install("GlobalAncova", version = "3.9")' && \
-    R -e 'BiocManager::install("impute", version = "3.9")' && \
-    R -e 'BiocManager::install("KEGGgraph", version = "3.9")' && \
-    R -e 'BiocManager::install("limma", version = "3.9")' && \
-    R -e 'BiocManager::install("pcaMethods", version = "3.9")' && \
-    R -e 'BiocManager::install("preprocessCore", version = "3.9")' && \
-    R -e 'BiocManager::install("Rgraphviz", version = "3.9")' && \
-    R -e 'BiocManager::install("siggenes", version = "3.9")' && \
-    R -e 'BiocManager::install("SSPA", version = "3.9")' && \
-    R -e 'BiocManager::install("sva", version = "3.9")'
+RUN R -e 'install.packages("XML");'
+RUN R -e 'install.packages("plotly");'
+RUN R -e 'install.packages("BiocManager"); BiocManager::install(version = "3.9", ask = FALSE)'
+RUN R -e 'BiocManager::install("genefilter", version = "3.9")'
+RUN R -e 'BiocManager::install("GlobalAncova", version = "3.9")'
+RUN R -e 'BiocManager::install("impute", version = "3.9")'
+RUN R -e 'BiocManager::install("KEGGgraph", version = "3.9")'
+RUN R -e 'BiocManager::install("limma", version = "3.9")'
+RUN R -e 'BiocManager::install("pcaMethods", version = "3.9")'
+RUN R -e 'BiocManager::install("preprocessCore", version = "3.9")'
+RUN R -e 'BiocManager::install("Rgraphviz", version = "3.9")'
+RUN R -e 'BiocManager::install("siggenes", version = "3.9")'
+RUN R -e 'BiocManager::install("SSPA", version = "3.9")'
+RUN R -e 'BiocManager::install("sva", version = "3.9")'
 
-RUN R -e 'install.packages("devtools"); library(devtools)'
-RUN R -e 'library(devtools); devtools::install_github("klutometis/roxygen")'
+RUN R -e 'install.packages("devtools");'
+RUN R -e 'install.packages("roxygen2")'
 RUN R -e 'BiocManager::install("impute", version = "3.9")'
 RUN R -e 'install.packages("missForest")'
 RUN R -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/mnormt/mnormt_1.5-6.tar.gz", repos=NULL, type="source")'
+RUN R -e 'require(devtools); install_version("mnormt", version = "1.5-6")'
 RUN R -e 'install.packages("psych")'
 RUN R -e 'install.packages("pROC")'
 ADD viime /viime
--- a/devops/Dockerfile.base
+++ b/devops/Dockerfile.base
@@ -33,6 +33,7 @@ RUN R -e 'library(devtools); devtools::install_github("klutometis/roxygen")'
 RUN R -e 'BiocManager::install("impute", version = "3.9")'
 RUN R -e 'install.packages("missForest")'
 RUN R -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/mnormt/mnormt_1.5-6.tar.gz", repos=NULL, type="source")'
-
+RUN R -e 'install.packages("psych")'
+RUN R -e 'install.packages("pROC")'
 ADD viime /viime
 RUN R -e 'setwd("/viime"); devtools::update_packages(devtools::dev_package_deps()$package, dependencies=NA, upgrade="never");' && rm -fr /viime
--- a/web/src/components/vis/GroupPredictionTile.vue
+++ b/web/src/components/vis/GroupPredictionTile.vue
@@ -55,12 +55,12 @@ export default {
       return this.dataset.groupLevels.map((level) => level.name);
     },
     metaboliteSourceOptions() {
-      const options = [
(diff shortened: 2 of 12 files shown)


mvandenburgh:
PxXO547I2W

jbeezley:
Is it the roc endpoint failing?  The test I added uses a column name with a space.  I might need to try with the table you are using.

mvandenburgh:
> Is it the roc endpoint failing? The test I added uses a column name with a space. I might need to try with the table you are using.

No, the roc endpoint looks up the column names in the validated measurement table but I was using the `column_header` field returned by the column's endpoint, so the server was reporting the `column` field as invalid. It's fixed now.

mvandenburgh:
Updated gif demo:
eJ6jWy67dU

jbeezley:
It is probably related to these lines: https://github.com/girder/viime/blob/roc/devops/viime/R/analyses.R#L163-L164.  I added `check.names=FALSE` because it was renaming the columns, but if none of the other endpoints are doing that, it could be a problem.  What do you think?  Would it be best to remove `check.names` from here, or add it to every other R method?

jeffbaumes:
This is looking great!

A few things I notice:

* My cursor seems to disappear when hovering over the plot, which does not seem helpful to me (I lose my place of where my mouse is).
* I sometimes see diagonal lines where ROC curve lines should be straight
* The analysis should not be called "ROC Curve" but instead something like "Group Prediction". Don't really mind what the endpoint is called but in the UI it should be changed.
* Metabolite chip overflow on the dropdown
* We should display the area under the curve, this serves as an overall metric of the prediction performance.
* The plot overflows to the right off the screen for me (seems to be using fixed size).

Some things to discuss with the group:

* How to select multiple metabolites: should we just use the global metabolite selection?
* How to specify the train/test split? This will likely need to be in a separate column of metadata.
* What train/validate modes do we want to use (cross-validate vs. explicit train/validate split)?

mvandenburgh:
> It is probably related to these lines: https://github.com/girder/viime/blob/roc/devops/viime/R/analyses.R#L163-L164. I added `check.names=FALSE` because it was renaming the columns, but if none of the other endpoints are doing that, it could be a problem. What do you think? Would it be best to remove `check.names` from here, or add it to every other R method?

When you say it's renaming the columns, you're referring to something like, for example, the metabolite named `3-Methoxy-4-hydroxyphenyl)ethylene glycol sulfate` being renamed to `X3.Methoxy.4.hydroxyphenyl.ethylene.glycol.sulfate` in the `roc.csv` dataset, right? I never noticed that the metabolites are renamed like that in some places.

I added `check.names=FALSE` to all the R methods and it still changed the column names around (unless I'm not rebuilding the docker image properly...). I think the renaming may have something to do with the serialization/deserialization in the `ValidatedMetaboliteTable` model.

jbeezley:
Did you change both of the `read.csv` calls in `impute.R`?  I just tried changing that and the column headers were unmodified in the validated file.

One consideration is if making this change will require migrating the database.  I suspect it won't because we always revalidate on page load.  I would say if it does require a migration, it's probably not worth the effort.

mvandenburgh:
> Did you change both of the `read.csv` calls in `impute.R`? I just tried changing that and the column headers were unmodified in the validated file.
> 
> One consideration is if making this change will require migrating the database. I suspect it won't because we always revalidate on page load. I would say if it does require a migration, it's probably not worth the effort.

Oh, I was changing them in `analyses.R`. I changed them in `impute.R` and the column headers are unmodified. 

I think we should make the change, I tested some datasets with the change and it worked without a migration.

jeffbaumes:
I've mocked up what I gathered from the meeting yesterday here: https://docs.google.com/presentation/d/1QdGB7o-hC37Hj_UIrCW14khPDOWEg0CQPCKmo1TMLmk/edit?usp=sharing

Here is the relevant diagram:

image

mvandenburgh:
- The ROC analysis opencpu function and endpoint now support selecting multiple metabolites and two groups. 

- Lilian's factor analysis R script is integrated into opencpu and a new endpoint is created for it.

- The web client UI is modified to account for these changes

- The "Metabolite Source" now allows filtering of metabolites based on PC values from factor analysis. There is also an option in the UI to allow the user to change the loading threshold used in the factor analysis.

- The metabolite chip overflow issue has been fixed.

web/src/components/vis/GroupPredictionTile.vue
```diff
@@ -0,0 +1,356 @@
...
+    // as a result of the user clicking the red 'X' next to its
+    // name in the 'Metabolites' autoselect
+    removeMetabolite(metabolite) {
+      this.metabolites.splice(this.metabolites.indexOf(metabolite), 1);
+      this.changePlotArgs({ columns: JSON.stringify(this.metabolites) });
+    },
+    async getFactors() {
+      if (this.metaboliteSourceOptions.length > 2) {
+        this.metaboliteSourceOptions.splice(2, this.metaboliteSourceOptions.length - 2);
+      }
```
dchiquit, line 119: Rather than having to do this stitching to keep the default options in the metabolite source list, why not use a computed variable? You're already storing `pcaData`, so you can simply make `metaboliteSourceOptions` computed with something like this:
```
metaboliteSourceOptions() {
  let options = [... default options ...];
  if (this.pcaData && this.pcaData.factors) {
    options.concat(this.pcaData.factor.map((factor) => { value: factor, text: `PC${factor}` }));
  }
}
```

web/src/components/vis/GroupPredictionTile.vue
```diff
@@ -0,0 +1,356 @@
...
+        auc: this.plot.data.auc[0],
+      };
+    },
+  },
+  watch: {
+    // get new factor analysis when threshold changes
+    threshold() {
+      this.getFactors();
+      this.metaboliteSource = 'all';
+    },
```
dchiquit, line 78: I'm not really sure what the significance of `changePlotArgs` is, but the only argument I don't see it called on anywhere is the loading threshold. Does it need to be called here?

web/src/components/vis/RocCurve.vue
```diff
@@ -0,0 +1,125 @@
...
+
+export default {
+  props: {
+    rocData: {
+      type: Object,
+      required: true,
+    },
+  },
+  mounted() {
+    // select('svg').remove(); // remove any existing graph
```
dchiquit, line 15: delete?

web/src/components/vis/RocCurve.vue
```diff
@@ -0,0 +1,125 @@
...
+import { select } from 'd3-selection';
+
+export default {
+  props: {
+    rocData: {
+      type: Object,
+      required: true,
+    },
+  },
+  mounted() {
```
dchiquit, line 14: It's kinda freaky to me that you can just throw the entire render function into the `mounted` callback. I assume it's fine because the plot is redrawn every time you change an argument, but I'm not sure how that behavior is happening here and not in other plots. It seems to work fine, though.

web/src/components/vis/RocCurve.vue
```diff
@@ -0,0 +1,125 @@
...
+
+    const { sensitivities } = this.rocData;
+    const { specificities } = this.rocData;
+
+    const data = [];
+
+    // Get the data to be plotted
+    sensitivities.forEach((sensitivity, i) => {
+      data.push({ sensitivity, specificity: 1 - specificities[i] });
+    });
```
dchiquit, line 69:
```suggestion
    const data = sensitivities.map((sensitivity, i) => { sensitivity, specificity: 1 - specificities[i] });
```

web/src/components/vis/GroupPredictionTile.vue
```diff
@@ -0,0 +1,356 @@
+<script>
```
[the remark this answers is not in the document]
mvandenburgh: I actually originally had it called `RocCurveTile.vue` but renamed it in https://github.com/girder/viime/pull/587/commits/e6b719b3d24ddd5f1a368a6e43031b18eb1e7230. I was following the convention that seemed to be set by `CorrelationTile.vue` being the name of the `ForceDirectedGraph.vue` component, which is that the tile wrapper shares a name similar to the name of the analysis in the UI, which is 'Group Prediction' in this case, while the component itself is named according to the actual visualization being shown, an ROC curve. Most of the other vis components happen to share the same name for these two. 

If you think it seems more reasonable to call it `RocCurveTile.vue` we can change it, it's mostly just a tile wrapper for `RocCurve.vue`.

web/src/components/vis/GroupPredictionTile.vue
mvandenburgh: Assuming I'm understanding how the `plotData` mixin works, calling `changePlotArgs` changes the arguments for the current analysis (ROC in this case), calls the corresponding analysis endpoint with the new arguments, and updates the `plot` object. 
The loading threshold is used for the factor analysis, which isn't an "analysis" in the same sense that the ROC analysis is, it's sort of a secondary analysis that is used by the ROC analysis/Group Prediction. It's only called once when the component is loaded initially, then every time the loading threshold changes.

web/src/components/vis/RocCurve.vue
```diff
@@ -0,0 +1,125 @@
...
+import { scaleLinear } from 'd3-scale';
+import { axisBottom, axisLeft } from 'd3-axis';
+import { line } from 'd3-shape';
+import { select } from 'd3-selection';
+
+export default {
+  props: {
+    rocData: {
+      type: Object,
+      required: true,
```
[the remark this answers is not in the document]
mvandenburgh: Agreed, that object is terrible in terms of code readability. Fixed in https://github.com/girder/viime/pull/587/commits/bfe433b003f9e60befa37a45900ea57eed285062.

web/src/components/vis/GroupPredictionTile.vue
mvandenburgh: Addressed in https://github.com/girder/viime/pull/587/commits/90fe6832a3dacef3ca4f9c92bba216192163d129.

web/src/components/vis/GroupPredictionTile.vue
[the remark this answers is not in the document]
dchiquit: If you've got a good reason I'm satisfied :+1:

devops/Dockerfile.base
```diff
@@ -32,6 +32,7 @@ RUN R -e 'install.packages("devtools"); library(devtools)'
 RUN R -e 'library(devtools); devtools::install_github("klutometis/roxygen")'
 RUN R -e 'BiocManager::install("impute", version = "3.9")'
 RUN R -e 'install.packages("missForest")'
+RUN R -e 'install.packages("https://cran.r-project.org/src/contrib/Archive/mnormt/mnormt_1.5-6.tar.gz", repos=NULL, type="source")'
```
jbeezley, line 35: Do you need me to push this up to dockerhub?

mvandenburgh: Yes, I just modified the base docker image to include the rest of the new dependencies for the opencpu server. Should make CI run faster since downloading those R packages takes a while.

jbeezley: I had to change a few things on the base image to get it to build.  I added the commit to this PR and pushed the changes to dockerhub.

waxlamp:
> @waxlamp may have more opinions about the D3 usage, but it looks good to me.

I will try to weigh in, but I'm fine with this having been merged already.

PR #589, opened 2020-07-14
Use watchEffect instead of update function

dchiquit:
One of the biggest pain points with using Vue and D3 is having an
`update` function behave reactively. Composition API offers a new
function, `watchEffect`, which will "Run a function immediately while
reactively tracking its dependencies, and re-run it whenever the
dependencies have changed."

Use `watchEffect` in `BoxPlot` and `LoadingsPlot`.

I'm not sure that it makes things that much clearer, but I thought I'd bring it up while we're refactoring.


web/src/components/vis/LoadingsPlot.vue
```diff
@@ -161,7 +161,7 @@ export default defineComponent({
     const axisX = computed(() => axisBottom(scaleX.value));
     const axisY = computed(() => axisLeft(scaleY.value));
 
-    function update() {
+    onMounted(() => watchEffect(() => {
```
subdavis, line 164: I was wondering in the previous PR how you'd use watcheffect with nullable refs.

I'm sort of convinced that the watchEffect should be pulled out of mounted and the refs should be allowed to be null before mount, and you should just have to check their value before using them.  

This seems to me like a concrete consequence of side-stepping typescript:

"Why did you put the watchEffect in onMounted?"
"Because the template refs could be null"

dchiquit: From https://github.com/girder/viime/pull/585#discussion_r457619012:
I personally thing the `onMounted(watchEffect(() => {...}));` is the least bad pattern. D3 requires the elements to be present before it can do it's thing, so it doesn't seem meaningful to be calling an update function prior to mounting. The null checks would basically be "Is the ref mounted yet? If not, abort the update", which can be completely circumvented by using `onMounted`.

waxlamp: If the refs can be null at some point in the lifecycle, what type does TS ascribe to them? If it's a "maybe" type, then it seems like we'd need to be doing `mainRef.value!` all over the place.

I think I agree with @dchiquit that this is the least friction way of writing this logic, though I also think the oddity of it deserves a code comment.

I would like to understand better what you mean by side-stepping typescript in this instance, @subdavis.

subdavis: I just mean that because the maybe-ness isn't captured formally in the type definition,  the watchEffect was moved into mounted in order to avoid an undefined check that a casual reader wouldn't realize is necessary.

This line seems to me to be compensating for the incorrect type.

If you move the watchEffect out of mounted and do the undefined check, then after mounted, when the ref is populated, watchEffect will fire again the way it's supposed to.  That seems to me to be the more rigorous option.

waxlamp: > I just mean that because the maybe-ness isn't captured formally in the type definition, the watchEffect was moved into mounted in order to avoid an undefined check that a casual reader wouldn't realize is necessary.
> 
> This line seems to me to be compensating for the incorrect type.

We should file an issue upstream discussing this rev-matching issue with the types (if you believe it's a formal enough issue to be captured thereby). I would expect a justification for the behavior rather than acceptance as a bug but it would be valuable to hear their take on it.

> 
> If you move the watchEffect out of mounted and do the undefined check, then after mounted, when the ref is populated, watchEffect will fire again the way it's supposed to. That seems to me to be the more rigorous option.

I would accept this approach as well. The code comment I requested would be targeted to the undefined check within that function.

waxlamp: > > I just mean that because the maybe-ness isn't captured formally in the type definition, the watchEffect was moved into mounted in order to avoid an undefined check that a casual reader wouldn't realize is necessary.
> > This line seems to me to be compensating for the incorrect type.
> 
> We should file an issue upstream discussing this rev-matching issue with the types (if you believe it's a formal enough issue to be captured thereby). I would expect a justification for the behavior rather than acceptance as a bug but it would be valuable to hear their take on it.
> 

@subdavis, do you think this is worth the trouble?

PR #591, opened 2020-07-17
Upload datasets test

JellyWang-3:
Initial click and upload datasets test.
It uploads 3 testing datasets to Viime.
test


e2e-test/src/tests/upload.test.js
```diff
@@ -0,0 +1,44 @@
...
+  it('upload one dataset and verify it', async () => {
+    await Promise.all([
+      page.goto(CLIENT_URL),
+      page.waitForNavigation({ waitUntil: 'networkidle0' }),
+    ]);
+    await expect(page).toClickXPath(vBtn('Your Datasets'));
+    const fileInputElementHandle = await page.waitForXPath("//input[@type='file']");
+    await fileInputElementHandle.uploadFile('../e2e-test/data/Cancer chemo muscle (NMR).csv');
+    const fileName = await page.waitForXPath("//div[@class='v-list__tile__title'][contains(text(),'Cancer_chemo_muscle_NMR.csv ')]");
+    expect(fileName).not.toBeNull();
```
JellyWang-3, line 15: I'm not sure about this approach. If there is the file that matches the name, then the return result won't be null.
I have tried the toHaveLength approach but it didn't work as I expected.
Selection_069

e2e-test/src/tests/upload.test.js
```diff
@@ -0,0 +1,44 @@
...
+    const datasetStatus = await page.waitForXPath("//div[@class='v-list__tile__title'][contains(text(),'Cancer_chemo_muscle_NMR.csv ')]//following::span[contains(text(),'Dataset processed with 1 validation failures')]");
+    expect(datasetStatus).not.toBeNull();
+  });
+  it('upload 3 dataset and verify it', async () => {
```
JellyWang-3, line 19: This might be redundant and I can remove it if needed.

e2e-test/src/tests/upload.test.js
dchiquit: [`page.waitForXPath`](https://devdocs.io/puppeteer/index#pagewaitforxpathxpath-options) returns the first element identified by an XPath, or throws an error if no matching element could be found within the timeout.
An error is thrown if it can't find the element, so it's enough to just call `waitForXPath`. The null check is technically dead code, since `waitForXPath` will throw an error instead of returnning null.
It will also make the assert more readable since you don't need the extra line, which is a nice bonus.

e2e-test/src/tests/upload.test.js
dchiquit: I don't think it's redundant.
However, it is quite hard to read. I recommend adding a blank line after loading the `Your Datasets` page, and also between each upload. That along with the above comment about removing the assert statements should make it much more legible.

PR #592, opened 2020-07-20
Delete tests

JellyWang-3:
I will remove the comments when it is good to go.


e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,66 @@
...
+    await expect(page).toClickXPath(vBtn('Your Datasets'));
+
+    const fileInputElementHandle = await page.waitForXPath("//input[@type='file']"); // $0.value == "" is true
+    await fileInputElementHandle.uploadFile('../e2e-test/data/Cancer chemo muscle (NMR).csv');
+    await page.waitForXPath("//div[@class='v-list__tile__title'][contains(text(),'Cancer_chemo_muscle_NMR.csv ')]");
+    await page.waitForXPath("//div[@class='v-list__tile__title'][contains(text(),'Cancer_chemo_muscle_NMR.csv ')]//following::span[contains(text(),'Dataset processed with 1 validation failures')]");
+
+    // click on 'X'
+    await page.waitForXPath("//div[@class='v-btn__content']//following::i[@class='v-icon mdi mdi-close theme--light']");
+    await expect(page).toClickXPath(vBtn("//i[@class='v-icon mdi mdi-close theme--light']"));
```
dchiquit, line 20: Firstly, these two XPaths are practically the same. The first is slightly more explicit, while the second uses the XPath helper some, but they should generate very similar XPaths.
Secondly, this should work the same and be much cleaner:
```
vBtn(vIcon('mdi-close'))
```

e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,66 @@
...
+
+    // make sure it is deleting one dataset
+    await page.waitForXPath("//div[@class='v-card__title headline'][contains(text(),'Really delete 1 dataset?')]");
```
dchiquit, line 23:
```suggestion
    await page.waitForXPath(vCard({ title: 'Really delete 1 dataset?' }));
```

e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,66 @@
...
+    // upload dataset 3
+    await fileInputElementHandle.uploadFile('../e2e-test/data/Cancer chemo serum (NMR).csv');
+    await page.waitForXPath("//div[@class='v-list__tile__title'][contains(text(),'Cancer_chemo_serum_NMR.csv ')]");
+    await page.waitForXPath("//div[@class='v-list__tile__title'][contains(text(),'Cancer_chemo_serum_NMR.csv ')]//following::span[contains(text(),'Dataset processed with 1 validation failures')]");
+
+    // click on 'clear all'
+    await expect(page).toClickXPath(vBtn('clear all'));
+
+    // make sure deleting 3 at the same time
+    await page.waitForXPath("//div[@class='v-card__title headline'][contains(text(),'Really delete 3 datasets?')]");
```
dchiquit, line 58:
```suggestion
    await page.waitForXPath(vCard({ title: 'Really delete 3 datasets?' }));
```

e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,66 @@
...
+    // click on 'X'
+    await page.waitForXPath("//div[@class='v-btn__content']//following::i[@class='v-icon mdi mdi-close theme--light']");
+    await expect(page).toClickXPath(vBtn("//i[@class='v-icon mdi mdi-close theme--light']"));
+
+    // make sure it is deleting one dataset
+    await page.waitForXPath("//div[@class='v-card__title headline'][contains(text(),'Really delete 1 dataset?')]");
+    await expect(page).toClickXPath(vBtn('Delete'));
+
+    // make sure it is deleted
+    await expect(page).not.toContainXPath("//div[@class='v-list__tile__title'][contains(text(),'Cancer_chemo_muscle_NMR.csv ')]");
```
dchiquit, line 27: I'm curious what happens if you run the upload test and then run this test. The upload test should leave behind the files that it uploaded, including this one. When this test runs it uploads the file again, but only deletes one, so after the deletion happens the file should still be on the page. Maybe this assertion is not doing what it is supposed to be doing?

The solution to this would be to use different files for different tests. For now it would be easiest to just have unique data files for every test. Eventually we might consider writing a helper function that takes a filename, appends some unique identifier to it, copies the original data file to the new longer file name, then uploads that. Seems like overkill for the tests we have now though.

e2e-test/src/tests/delete.test.js
dchiquit: Also, if there are any leftover datasets, clearing all might give a number larger than 3. This test should probably be clicking this button at the beginning of the test to make sure there are no extra files hanging around to interfere.

In fact, that might be a handy utility function that could be used in any tests that don't want extra files hanging around. What do you think?

e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('upload 3 datasets and then clear all', async () => {
```
dchiquit, line 30: The `upload 3 datasets` part isn't really what's being tested, it's just a setup step. I would just call this test `clear all` as that is what is being tested here.
Similar comments apply to other tests as well, I didn't notice earlier.

e2e-test/src/tests/delete.test.js
JellyWang-3: Yeah I agree. So write a function an reuse it for different tests to make sure all the files are cleared?

e2e-test/src/tests/delete.test.js
JellyWang-3: I did not realize that the files are left behind that is probably why sometimes the CI tests are failing. Because there are more than 3 datasets already. I will try to use unique datasets for different tests. So far I was able to download the three from https://viime.org/#/ are there any more datasets that I can use? Where can I find more?

dchiquit: The easiest thing would be to just take those datasets and rename them. The data will be the same, but they will be considered unique by VIIME.

If you ever do need different data to test a feature or something, you can try the [metaboanalyst samples](https://www.metaboanalyst.ca/MetaboAnalyst/upload/StatUploadView.xhtml). They're harder to download but there are more of them.

e2e-test/src/tests/delete.test.js
dchiquit: Yep, specifically for tests that need to start with no existing files (like this one). Most tests probably won't use it.

JellyWang-3:
I wrote a very simple function in util.js and tried to export it to the delete test.
Maybe it's the way I am calling/using this function but it would give me this error when I ran the tests.

Selection_073

For now I just put the actual code in delete.test.js because it only has 2 lines and I could not find a way to fix the error above.

I do have a question about should I double check if the upload file area is cleared after all the datasets are delete or no? It might be redundant because I already checked the file upload area at first and then verified the files uploaded are all deleted later.

dchiquit:
That error looks like it has to do with an asynchronous JS function call. If a function is defined as asynchronous, for example async helper() {...}, it will always run asynchronously. If you just call helper(), your code will trigger helper to run eventually, but will then immediately move on and execute the next line. The only way to make helper behave like a normal function is to do await helper(), which will wait for the helper thread to finish before resuming the original thread.

Anyway, I've seen errors like that when I've called async functions and forgotten to include the await. I expect the function to run immediately, but instead the test proceeds and finishes before the async function can be invoked.
Here's what I think happened. You defined some `async helper() {...}` in `util.js`, which is great. Near the beginning of the test, you call `helper()`. `helper` is async, so it doesn't run immediately; it just sits in the event loop until the test has finished. Now the test is over and the browser is closed, but `helper` still has to run. It tries to use the browser, which is closed, so it throws a `Target closed` exception. There's also a warning there, `Jest did not exit one second after the test run has completed`. That's a huge red flag that you have leftover async function calls hanging around.

e2e-test/src/util.js
```diff
@@ -1,2 +1,7 @@
-/* eslint-disable import/prefer-default-export */
 export const { CLIENT_URL } = process.env;
+
+export async function checkNoFile() {
+  // check if there are any other files hanging around
+  const uploadFileList = await expect(page).toClickXPath("//div[@class='v-list upload-list theme--light']");
```
dchiquit, line 5: This doesn't seem to do what it advertises that it does. All it does is click on the upload list, which isn't even meaningful; I think puppeteer would just click in the middle of the upload list element, which will be somewhere on some element in the list, if it exists.
If you're checking that no files exist, you should return a boolean.

e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,65 @@
...
+    const fileInputElementHandle = await page.waitForXPath("//input[@type='file']");
+    await fileInputElementHandle.uploadFile('../e2e-test/data/deleteTest.csv');
+    await page.waitForXPath(vListTile({ title: 'deleteTest.csv ' }));
+    await page.waitForXPath(vListTile({ title: 'deleteTest.csv ', content: 'Dataset ready for analysis' }));
+
+    await expect(page).toClickXPath(vBtn(vIcon('mdi-close')));
+    await page.waitForXPath(vCard({ headline: 'Really delete 1 dataset?' }));
+    await expect(page).toClickXPath(vBtn('Delete'));
+
+    const previouslyUploadedFile = await expect(page).toContainXPath(vListTile({ title: 'deleteTest.csv ' }));
```
JellyWang-3, line 30: To verify one specific file that has been deleted

e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,65 @@
...
+    await fileInputElementHandle.uploadFile('../e2e-test/data/deleteTest3.csv');
+    await page.waitForXPath(vListTile({ title: 'deleteTest3.csv ' }));
+    await page.waitForXPath(vListTile({ title: 'deleteTest3.csv ', content: 'Dataset processed with 1 validation failures' }));
+
+    await expect(page).toClickXPath(vBtn('clear all'));
+
+    await page.waitForXPath(vCard({ headline: 'Really delete 3 datasets?' }));
+    await expect(page).toClickXPath(vBtn('Delete'));
+
+    await allCleared();
```
JellyWang-3, line 63: util function to verify all the uploaded files are cleared

e2e-test/src/util.js
```diff
@@ -1,2 +1,18 @@
-/* eslint-disable import/prefer-default-export */
+import { vBtn, vList } from 'jest-puppeteer-vuetify';
+
 export const { CLIENT_URL } = process.env;
+
+export async function deleteAllDatasets() {
+  // firstUploadedItem returns a uploaded file node list
```
JellyWang-3, line 6: I can take out this comment if needed

e2e-test/src/tests/delete.test.js
```diff
@@ -0,0 +1,65 @@
...
+
+    // todo: make a function for upload file(s)
+    const fileInputElementHandle = await page.waitForXPath("//input[@type='file']");
+    await fileInputElementHandle.uploadFile('../e2e-test/data/deleteTest.csv');
+    await page.waitForXPath(vListTile({ title: 'deleteTest.csv ' }));
+    await page.waitForXPath(vListTile({ title: 'deleteTest.csv ', content: 'Dataset ready for analysis' }));
+
+    await expect(page).toClickXPath(vBtn(vIcon('mdi-close')));
+    await page.waitForXPath(vCard({ headline: 'Really delete 1 dataset?' }));
+    await expect(page).toClickXPath(vBtn('Delete'));
```
dchiquit, line 28: This code isn't working, and neither is the assert. As a general rule you should always deliberately break your tests once you've written them to make sure that they would correctly fail when something breaks.
This code is running too fast; I think it's deleting the item before it's been saved to the database, so the item is still there with a little `FATAL ERROR` chip attached to it. Adding a `await new Promise(resolve => setTimeout(resolve, 1000));` will probably fix it.

e2e-test/src/util.js
```diff
@@ -1,2 +1,18 @@
...
+  const firstUploadedItem = await page.$x(vList({ content: 'upload-list theme--light' }));
+  if (firstUploadedItem.length !== 0) {
+    await expect(page).toClickXPath(vBtn('clear all'));
+    await page.waitForXPath(vBtn('Delete'));
+    await expect(page).toClickXPath(vBtn('Delete'));
+  }
+}
+
+export async function allCleared() {
+  const firstUploadedItem = await page.$x(vList({ content: 'upload-list theme--light' }));
```
dchiquit, line 16: This wasn't working. Almost though:
```suggestion
  const firstUploadedItem = await page.$x(vList({ cssClass: 'upload-list theme--light' }));
```
Also you may need to add a second wait to ensure that everything has time to be removed.

e2e-test/src/tests/delete.test.js
```diff
@@ -23,12 +24,13 @@ describe('delete dataset', () => {
...
     await page.waitForXPath(vListTile({ title: 'deleteTest.csv ', content: 'Dataset ready for analysis' }));
 
+    await new Promise((resolve) => setTimeout(resolve, 1000));
     await expect(page).toClickXPath(vBtn(vIcon('mdi-close')));
     await page.waitForXPath(vCard({ headline: 'Really delete 1 dataset?' }));
     await expect(page).toClickXPath(vBtn('Delete'));
 
-    const previouslyUploadedFile = await expect(page).toContainXPath(vListTile({ title: 'deleteTest.csv ' }));
-    await expect(previouslyUploadedFile).toBeUndefined();
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
JellyWang-3, line 32: Wait for the data set to be deleted then verify.

e2e-test/src/tests/delete.test.js
```diff
@@ -55,11 +57,13 @@ describe('delete dataset', () => {
...
     await page.waitForXPath(vListTile({ title: 'deleteTest3.csv ', content: 'Dataset processed with 1 validation failures' }));
 
+    await new Promise((resolve) => setTimeout(resolve, 1000));
     await expect(page).toClickXPath(vBtn('clear all'));
-
     await page.waitForXPath(vCard({ headline: 'Really delete 3 datasets?' }));
     await expect(page).toClickXPath(vBtn('Delete'));
 
-    await allCleared();
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
JellyWang-3, line 65: Wait to make sure it's all cleared then verify.

e2e-test/src/tests/delete.test.js
```diff
@@ -23,12 +24,13 @@ describe('delete dataset', () => {
     await page.waitForXPath(vListTile({ title: 'deleteTest.csv ' }));
     await page.waitForXPath(vListTile({ title: 'deleteTest.csv ', content: 'Dataset ready for analysis' }));
 
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
JellyWang-3, line 27: Wait for data set to be uploaded

e2e-test/src/tests/delete.test.js
```diff
@@ -55,11 +57,13 @@ describe('delete dataset', () => {
     await page.waitForXPath(vListTile({ title: 'deleteTest3.csv ' }));
     await page.waitForXPath(vListTile({ title: 'deleteTest3.csv ', content: 'Dataset processed with 1 validation failures' }));
 
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
JellyWang-3, line 60: Wait for data sets to be all uploaded before deletion.

e2e-test/src/tests/delete.test.js
dchiquit: If you feel like you need to explain yourself in the PR, you probably also need to explain yourself to anyone reading the code. I think it's generally a good idea to include a comment explaining why a sleep is necessary.

PR #601, opened 2020-07-29
Relabel test

JellyWang-3:
I have broken down a that long user story into 2 pieces of tests.
I could use some suggestions on the amount of comments/ should I change some comments to make things more clear.


e2e-test/src/tests/cleanUp.test.js
```diff
@@ -0,0 +1,106 @@
+import {
+  vBtn,
+  vListTile,
+  vIcon,
+} from 'jest-puppeteer-vuetify';
+import { CLIENT_URL } from '../util';
+
+
+describe('relabel dataset', () => {
+  it('make dataset ready for analysis with 2 clicks', async () => {
```
JellyWang-3, line 10: There might be a better way to name this

e2e-test/src/tests/cleanUp.test.js
```diff
@@ -0,0 +1,106 @@
...
+    await expect(page).toClickXPath("//div[@class='column-header-cell type-masked mdi mdi-eye-off']");
+
+    // select the masked column and mark it as a 'Group'
+    await expect(page).toClickXPath("//input[@value='group']");
+
+    //  verify the dataset is ready to process again
+    await page.waitForXPath(vListTile({ title: 'cleanUpTest1.csv' }) + vIcon({ cssClass: 'mdi mdi-check theme--light success--text' }));
+  });
+
+  it('fix selecting wrong data type for file name', async () => {
```
JellyWang-3, line 69: There might be a better way to name this

e2e-test/src/tests/cleanUp.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+    // select the third column(column 'C') and mark it as a 'Group'
+    await expect(page).toClickXPath("//div[@class='column-header-cell type-measurement'][contains(text(), 'C')]");
+    await expect(page).toClickXPath("//input[@value='group']");
+
+    //  verify the dataset is ready to process but needs to be more explict
+    await page.waitForXPath(vListTile({ title: 'cleanUpTest.csv' }) + vIcon({ cssClass: 'mdi mdi-check theme--light success--text' }));
+  });
+
+  it('mask all and unmask', async () => {
```
JellyWang-3, line 34: There might be a better way to name this

e2e-test/src/tests/cleanUp.test.js
```diff
@@ -0,0 +1,106 @@
...
+    await expect(page).toClickXPath(vListTile({ title: 'cleanUpTest.csv ', content: 'Dataset processed with 1 validation failures' }) + vBtn('View Data'));
+
+    //  verify the dataset is failing
+    await page.waitForXPath(vListTile({ title: 'cleanUpTest.csv' }) + vIcon({ cssClass: 'mdi mdi-alert theme--light warning--text' }));
+
+    // select the third column(column 'C') and mark it as a 'Group'
+    await expect(page).toClickXPath("//div[@class='column-header-cell type-measurement'][contains(text(), 'C')]");
+    await expect(page).toClickXPath("//input[@value='group']");
+
+    //  verify the dataset is ready to process but needs to be more explict
```
dchiquit, line 30: What does `but needs to be more explict` mean? I can see how this would verify that the dataset is ready to process but I don't see anything here about missing info.
Also, typo

e2e-test/src/tests/cleanUp.test.js
dchiquit: How about `mark as group`? That's the relabeling action that is being done in this test.

e2e-test/src/tests/cleanUp.test.js
dchiquit: I think this name is fine :+1:

e2e-test/src/tests/cleanUp.test.js
```diff
@@ -0,0 +1,106 @@
...
+} from 'jest-puppeteer-vuetify';
+import { CLIENT_URL } from '../util';
+
+
+describe('relabel dataset', () => {
+  it('make dataset ready for analysis with 2 clicks', async () => {
+    await Promise.all([
+      page.goto(CLIENT_URL),
+      page.waitForNavigation({ waitUntil: 'networkidle0' }),
+    ]);
```
dchiquit, line 14: Now that I'm looking at it, I think this step might be unnecessary. This snippet is also being run in a `beforeAll` in `jest.setup.js`, which should run it before every test. I'm curious if removing it breaks anything. I should have caught that sooner.

e2e-test/src/tests/cleanUp.test.js
```diff
@@ -0,0 +1,106 @@
...
+    const fileInputElementHandle = await page.waitForXPath("//input[@type='file']");
+    await fileInputElementHandle.uploadFile('../e2e-test/data/cleanUpTest1.csv');
+    await page.waitForXPath(vListTile({ title: 'cleanUpTest1.csv ', content: 'Dataset processed with 1 validation failures' }));
+
+    await expect(page).toClickXPath(vListTile({ title: 'cleanUpTest1.csv ', content: 'Dataset processed with 1 validation failures' }) + vBtn('View Data'));
+
+    //  verify the dataset is failing
+    await page.waitForXPath(vListTile({ title: 'cleanUpTest1.csv' }) + vIcon({ cssClass: 'mdi mdi-alert theme--light warning--text' }));
+
+    // verify the dataset gets an error
```
dchiquit, line 50:
```suggestion
    // click the error message to get fix recommendations
```

e2e-test/src/tests/cleanUp.test.js
dchiquit: `mark as metabolite + metadata`?

dchiquit: By the way, while it's generally a good idea to have separate tests for separate functions, this is a case where it's just so much easier to test the `Metabolite` and `Metadata` buttons in the same test.

e2e-test/src/tests/cleanUp.test.js
JellyWang-3: That was some personal notes that I forgot to take out.
It should just be: `//  verify the dataset is ready to process`

PR #602, opened 2020-07-29
Typescript volcano plot

### LINKED ISSUE #595: Typescript Volcano Plot

mvandenburgh:
Resolves #595.

### DIFF
--- a/web/src/components/vis/VolcanoPlot.vue
+++ b/web/src/components/vis/VolcanoPlot.vue
@@ -34,7 +34,7 @@ export default defineComponent({
   props: {
     rows: {
       type: Array as PropType<Row[]>,
-      default: () => [],
+      required: true,
     },
     minFoldChange: {
       type: Number,
@@ -47,7 +47,8 @@ export default defineComponent({
   },
 
   setup(props) {
-    const mainRef = ref(document.createElement('div'));
+    const svgRef = ref() as Ref<HTMLElement>;
+    const mainRef = ref() as Ref<HTMLElement>;
 
     const width = ref(800);
     const height = computed(() => (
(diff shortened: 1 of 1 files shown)


web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,160 @@ export default {
...
-      | {{ylabel}}
+<template>
+  <div
+    id="mainRef"
+    ref="mainRef"
+    v-resize:throttle="onResize"
+    class="main"
+  >
+    <svg
+      ref="svg"
```
dchiquit, line 184: This ref seems unused to me.

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -1,17 +1,40 @@
...
+const xlabel = 'log2(Fold Change)';
+const ylabel = '-log10(p-value)';
 
+export default defineComponent({
   props: {
     rows: {
-      type: Array, // {pValue: number, log2FoldChange: number, name: string, color?: string}[]
-      required: true,
+      type: Array as PropType<Row[]>,
+      default: () => [],
```
dchiquit, line 37: Was there a problem with making `rows` a required prop?

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,160 @@ export default {
...
+    const axisX = computed(() => axisBottom(scaleX.value));
+    const axisY = computed(() => axisLeft(scaleY.value));
+
+    function onResize() {
+      const bb = mainRef.value.getBoundingClientRect();
+      width.value = bb.width;
+    }
+    onMounted(() => onResize());
+    watchEffect(() => {
+      const svg = select('svg');
```
dchiquit, line 105: This is interesting. I ran into problems when I defined an `svgRef` where calling `select(svgRef.value)` in `watchEffect` before the component was mounted would cause errors because the ref wasn't bound yet. I solved it by doing `onMounted(() => watchEffect(...));`. You use `select('svg') here, which I assume somehow works prior to component mounting? I'm surprised this works as is.

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,160 @@ export default {
...
+        // @ts-ignore
+        .attr('opacity', (d) => (Math.abs(d.x) >= props.minFoldChange && d.y >= props.minLogP ? 1 : 0.5))
+        // @ts-ignore
+        .attr('cx', (d) => _scaleX(d.x))
+        // @ts-ignore
+        .attr('cy', (d) => _scaleY(d.y))
+        // @ts-ignore
         .style('fill', (d) => d.color)
         .select('title')
+        // @ts-ignore
```
dchiquit, line 130: This is a depressing number of `@ts-ignore`. Is there a way to resolve whatever it's whining about, or to ignore a whole block?

web/src/components/vis/VolcanoPlot.vue
mvandenburgh: I changed it to use an `svgRef` variable inside `onMounted()` like you have in the Scree Plot to make things more uniform.

web/src/components/vis/VolcanoPlot.vue
mvandenburgh: It turns out only one `@ts-ignore` was needed at the start of the method chaining. I worked from the bottom of the file up when I was fixing those typescript errors so I missed that.

web/src/components/vis/VolcanoPlot.vue
mvandenburgh: No, fixed in https://github.com/girder/viime/pull/602/commits/8b890c887770c9a09fcaec80018901a3f408ced3

web/src/components/vis/VolcanoPlot.vue
mvandenburgh: Addressed in https://github.com/girder/viime/pull/602/commits/8b890c887770c9a09fcaec80018901a3f408ced3

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,146 @@ export default {
...
+        const bb = mainRef.value.getBoundingClientRect();
+        width.value = bb.width;
+
+        const svg = select(svgRef.value);
+        axisPlot(svg, axisX.value, axisY.value);
+
+        const _scaleX = scaleX.value;
+        const _scaleY = scaleY.value;
+
+        // @ts-ignore
```
waxlamp, line 110: From the TS docs:

> A // @ts-ignore comment suppresses all errors that originate on the following line. It is recommended practice to have the remainder of the comment following @ts-ignore explain which error is being suppressed.

Could you add a comment in this format explaining why this is needed? (I assume it's because D3 has bad types defined for methods like `select()` and `selectAll()`.)

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,146 @@ export default {
...
+      watchEffect(() => {
+        // Recalculate width/height
+        const bb = mainRef.value.getBoundingClientRect();
+        width.value = bb.width;
+
+        const svg = select(svgRef.value);
+        axisPlot(svg, axisX.value, axisY.value);
+
+        const _scaleX = scaleX.value;
+        const _scaleY = scaleY.value;
```
waxlamp, line 108: As a style matter, I think it would be better to have a more explicit name for these, rather than use the prefix underscore (which can have a special sort of meaning in JS code in general, different from what you're aiming for here). `scaleXVal` and `scaleYVal` could be better names. Or you could consider simply writing `.value` out at the points of use; that would keep it very explicit throughout the code that you're making use of a ref.

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,146 @@ export default {
...
+          .attr('y2', height.value - margin.top - margin.bottom);
+
+        svg.select('.plot').selectAll('line.y-threshold')
+          .data([1])
+          .join('line')
+          .attr('class', 'y-threshold')
+          .style('stroke', 'black')
+          .style('stroke-width', 0.5)
+          .attr('x1', 0)
+          .attr('y1', () => _scaleY(props.minLogP))
```
waxlamp, line 143: This is a function computing a static value; it seems like you could just remove the function call altogether:

```suggestion
          .attr('y1', _scaleY(props.minLogP))
```

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,146 @@ export default {
...
+          .attr('x2', width.value - margin.left - margin.right)
+          .attr('y2', () => _scaleY(props.minLogP));
```
waxlamp, line 145: Same situation as above:

```suggestion
          .attr('y2', _scaleY(props.minLogP));
```

mvandenburgh:
@waxlamp I addressed your comments in https://github.com/girder/viime/pull/602/commits/67d112c903919068c1d5043a22bb531f10235fce.

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -1,16 +1,39 @@
...
+}
 
-export default {
-  mixins: [
-    axisPlot,
-  ],
+interface TransformedRow extends Row {
+  x: number;
+  y: number;
+}
```
subdavis, line 20: I like the use of `extends` to describe what additional properties decorate the regular data.

web/src/components/vis/VolcanoPlot.vue
```diff
@@ -23,106 +46,143 @@ export default {
...
+    const scaleY = computed(() => scaleLinear()
+      .domain(yrange.value)
+      .range([dheight.value, 0]));
+
+    const axisX = computed(() => axisBottom(scaleX.value));
+    const axisY = computed(() => axisLeft(scaleY.value));
+
+    onMounted(() => {
+      watchEffect(() => {
+        // Recalculate width/height
```
subdavis, line 100: Even more than with #604 I think it'd be good to hoist dependencies to the top of this function. 

D3 style makes it hard (for me) to figure out where looping is happening, and reading from a reactive variable incurs a cost that can be amplified to noticeable slowness in a loop, so you might even get better performance by taking the read penalty once at the top.

mvandenburgh: Agreed, I moved the dependencies to the top of the `watchEffect` function.

PR #604, opened 2020-07-30
Add usePlotData and convert ScreePlotTile to TS

dchiquit:
This is mostly about converting the `plotData.js` mixin to `usePlotData.ts`. I think it's much cleaner but I'm open to suggestions, especially about the watchEffect.

I did not convert the pug in `ScreePlotTile` to HTML because I wasn't sure what to do about the `include`.

### DIFF
--- a/web/src/components/vis/ScreePlotTile.vue
+++ b/web/src/components/vis/ScreePlotTile.vue
@@ -1,46 +1,58 @@
-<script>
+<script lang="ts">
+import { defineComponent, computed, ref } from '@vue/composition-api';
 import ScreePlot from '@/components/vis/ScreePlot.vue';
 import VisTile from '@/components/vis/VisTile.vue';
-import plotData from './mixins/plotData';
+import usePlotData from './use/usePlotData';
 
-export default {
-  components: {
-    ScreePlot,
-    VisTile,
-  },
-
-  mixins: [plotData('pca')],
-
+export default defineComponent({
   props: {
     id: {
-      required: true,
       type: String,
+      required: true,
     },
   },
 
-  data() {
+  components: { ScreePlot, VisTile },
+
+  setup(props) {
+    const id = computed(() => props.id);
+    const showCutoffs = ref(true);
+    const numComponentsText = ref('10');
+    const numComponents = computed(() => Number.parseInt(numComponentsText.value, 10));
+
+    const { plot } = usePlotData(id, 'pca');
+
     return {
-      numComponentsText: '10',
-      showCutoffs: true,
+      numComponentsText,
+      numComponents,
+      showCutoffs,
+      plot,
     };
   },
-
-  computed: {
-    numComponents() {
-      return Number.parseInt(this.numComponentsText, 10);
-    },
-  },
-};
+});
 </script>
 
 <template lang="pug">
-vis-tile(v-if="plot", title="PCA Scree Plot", :loading="plot.loading", svg-download)
+vis-tile(
+  v-if="plot",
+  title="PCA Scree Plot",
+  :loading="plot.loading",
+  svg-download
+)
   scree-plot(
-      :eigenvalues="getPlotDataProperty('sdev')",
-      :num-components="numComponents",
-      :show-cutoffs="showCutoffs")
+    v-if="plot.data",
+    :eigenvalues="plot.data.sdev",
+    :num-components="numComponents",
+    :show-cutoffs="showCutoffs"
+  )
   template(v-slot:controls)
-    v-menu(bottom, offset-y, left, :min-width="150", :close-on-content-click="false")
+    v-menu(
+      bottom,
+      offset-y,
+      left,
+      :min-width="150",
+      :close-on-content-click="false"
+    )
       template(v-slot:activator="{ on }")
         v-btn(v-on="on", icon)
           v-icon.mdi.mdi-dots-vertical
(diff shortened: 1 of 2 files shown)


web/src/components/vis/ScreePlotTile.vue
```diff
@@ -1,42 +1,42 @@
...
-import plotData from './mixins/plotData';
-
-export default {
-  components: {
-    ScreePlot,
-    VisTile,
-  },
-
-  mixins: [plotData('pca')],
+import usePlotData from './use/usePlotData';
```
waxlamp, line 5: Is this `use` directory a recommended best practice for storing mixins?

web/src/components/vis/use/usePlotData.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  computed, Ref, ref, onActivated, onDeactivated, watchEffect,
+} from '@vue/composition-api';
+import store from '../../../store';
+import { LOAD_PLOT } from '../../../store/actions.type';
+import { SET_PLOT } from '../../../store/mutations.type';
+
+export default function useDataPlot(id: Ref<string>, plotName: string) {
```
waxlamp, line 8: I believe this should be `usePlotData`, right?

```suggestion
export default function usePlotData(id: Ref<string>, plotName: string) {
```

web/src/components/vis/use/usePlotData.ts
```diff
@@ -0,0 +1,41 @@
...
+      name: plotName,
+      obj: { args: { ...plot.value.args, ...args } },
+    });
+  }
+
+  onActivated(() => { active.value = true; });
+  onDeactivated(() => { active.value = false; });
+  watchEffect(() => {
+    // TODO I think this is cleaner, but I'm not sure if will register properly.
+    // It needs to watch for changes to plot, active, and id
```
waxlamp, line 26: Any evidence that this won't work? On my reading of the `watchEffect` docs, this looks like it will be reactive on `plot`, `active`, and `id`.

web/src/components/vis/ScreePlotTile.vue
```diff
@@ -1,42 +1,42 @@
...
       type: String,
+      required: true,
     },
   },
 
-  data() {
+  components: { ScreePlot, VisTile },
+
+  setup(props) {
+    const id = computed(() => props.id);
```
subdavis, line 18: Is the point of this to protect the `usePlotData` function by wrapping the prop in a readonly ref?

You'll probably want to mark this as a TODO, since in Vue 3 it will be fine to pass `props.id` directly to `usePlotData`

web/src/components/vis/use/usePlotData.ts
subdavis, line 26: In general, code >>> comments.  I'd rather have clearly established dependencies at the top of a watchEffect() than a probably wrong comment about what the watchEffect function used to depend on.

I'm rather fond of writing computed dependencies like this:

```
computedThing() {
  // establish dependencies
  const { foo, bar, baz } = this;
  // then, use of "this" is prohibited by convention.
}
```

This actually has a performance penalty if your logical branching is unbalanced: For example if foo is always false, but inside some foo=true conditional you establish the bar dependency, and bar gets updated 100 times per second, this would be a real problem.   In most cases, though, the performance-vs-readability trade-off is well worth using the above style.

I don't know what the equivalent watchEffect style would be, but in my head it looks something like this:

```suggestion
  const plotVal = plot.value;
  const activeVal = active.value;
  const idVal = id.value;
```

I'd like feedback from others here.

waxlamp: I second the practice of pulling out reactive dependencies into local variables in the function "header". I try to do this using destructuring assignment (as @subdavis demoed above) whenever possible. Since that's not possible here, a sequence of `const` variable declarations specifying which refs are used works just as well.

mvandenburgh: I agree with this as well, putting the reactive dependencies at the top of the function makes it much more readable and easier to figure out where/why UI changes occur. As far as syntax for `watchEffect` dependencies, I think the way @subdavis described is the best - at least I can't think of a better solution. 

I do like how clean/self-documenting the destructuring assignment with `this` for computed properties is, I haven't seen that before.

web/src/components/vis/ScreePlotTile.vue
dchiquit: I'm copying Brandon's `useAxisPlot`, I don't know if it's a best practice.

web/src/components/vis/ScreePlotTile.vue
dchiquit: I was basically trying to make `props.id` reactive. `props` is reactive, but `props.id` is just a string, so if you navigate to a different dataset `props.id` will update, but that won't propagate from here to `usePlotData`.
If there's a better way to make it reactive that would be awesome, this feels gross.

web/src/components/vis/use/usePlotData.ts
dchiquit: In my defense I meant for the comment to be more of an RFC than a permanent feature.
I like the pseudo-destructuring pattern, it feels like the best solution to me (although I named things `plotValue` instead of `plotVal`).

waxlamp: > In my defense

I took it to be just a comment, rather than something permanent :smile: 

That's also why I asked for it to be removed before merge (once you were happy with the function on the whole).

PR #605, opened 2020-08-03
Add error handling to ANOVA

### LINKED ISSUE #599: No ANOVA results looks like error condition
via @joannawolthuis:
> Furthermore it seems I have no ANOVA results - I would add something like 'no significant hits found' as now I just have an empty screen and do not know if that is due to an error or due to no hits being found.

(split from #597)

mvandenburgh:
It's possible for the ANOVA test to fail on a dataset; currently this results in the R script crashing, which causes OpenCPU to return an error, which in turn triggers a sentry error in production. This PR will add error-handling to gracefully handle these errors on the server and display an appropriate error message in the web client. 

Resolves #599.

### DIFF
--- a/devops/viime/R/analyses.R
+++ b/devops/viime/R/analyses.R
@@ -77,7 +77,21 @@ anova_tukey_adjustment <- function(measurements, groups, log_transformed=FALSE) 
   OUT <- NULL
   for (n in colnames(Metab)){
     mod <- lm(Metab[[n]] ~ Group)
-    a <- Anova(mod, type="III")
+
+    # catches any possible errors thrown by ANOVA and returns them
+    a <- tryCatch(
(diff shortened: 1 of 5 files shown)


mvandenburgh:
Here's a demo of the ANOVA failure now -

o0SQTrse9t

Right now it just displays the error message returned by R inside the empty table. This PR also adds support for showing a custom error message in the `MetaboliteTable` component, instead of the standard vuetify `No data available` text. Not sure if that may be useful in the future for other components, or if it just adds more complexity and/or would be better to implement this differently in the UI.

mvandenburgh:
@subdavis I addressed your comments in https://github.com/girder/viime/pull/605/commits/34340aca899047bc7631a25e98f71a95169170ed

PR #612, opened 2020-08-05
Transform, anova, and download tests

JellyWang-3:
The test cases are for testing buttons on analyze/transform table. 
The test cases would check if the values in tables have being modified after certain buttons being clicked/ check boxes are selected.
Finding the xpath was not easy and there might be better way to specify some xpath.


e2e-test/src/tests/analyze.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+    // click transform table
+    await expect(page).toClickXPath(vListTile({ title: 'Transform Table' }));
+
+    // change normalize to 'Min Max'
+    await expect(page).toClickXPath("//label[@class='v-label theme--light'][contains(text(), 'Min Max')]");
+    await expect(page).toClickXPath(vListTile({ title: 'Download Data' }));
+
+    // verify the value has been changed: hard coded but there might be a better way to do that
+    await page.waitForXPath("//div[@class='vue-recycle-scroller__item-view'][1]//div[@class='column']/div[@class='cell type-sample'][1][contains(text(), '0.249')]");
```
JellyWang-3, line 20: I could not think of a better way to verify the value has changed except hard code the value to be changed to an expected number.

e2e-test/src/tests/analyze.test.js
```diff
@@ -0,0 +1,106 @@
...
+    // click download data to verify the change
+    await expect(page).toClickXPath(vListTile({ title: 'Download Data' }));
+    await page.waitForXPath("//div[@class='vue-recycle-scroller__item-view'][1]//div[@class='column']/div[@class='cell type-sample'][1][contains(text(), '-0.603')]");
+
+    await expect(page).toClickXPath(vListTile({ title: 'Transform Table' }));
+
+    // change scale to 'Pareto Scaling'
+    await expect(page).toClickXPath("//label[@class='v-label theme--light'][contains(text(), 'Pareto Scaling')]");
+    await expect(page).toClickXPath(vListTile({ title: 'Download Data' }));
+    await page.waitForXPath("//div[@class='vue-recycle-scroller__item-view'][1]//div[@class='column']/div[@class='cell type-sample'][1][contains(text(), '0.393')]");
```
JellyWang-3, line 35: I don't know how to verify the plot has changed. I could take screenshot but I don't what would I compare that to.
I have no idea how to verify that the graph has changed.

e2e-test/src/tests/analyze.test.js
```diff
@@ -0,0 +1,106 @@
...
+    await expect(page).toClickXPath(vListTile({ title: 'Analyze Table' }));
+
+    // click ANOVA
+    await expect(page).toClickXPath(vListTile({ title: 'ANOVA' }));
+
+    // select some metabolites from ANOVA by adding the highlighted Metabolites to the selected set
+    /* eslint prefer-template: "off" */
+    await expect(page).toClickXPath("//*[text()='C26 - C26.Fol']" + vBtn({ title: 'Adds the highlighted Metabolites to the selected set' }));
+    selectedMetabolites = await page.$x("//div[@class='v-input v-input--selection-controls v-input--checkbox v-input--hide-details v-input--is-label-active v-input--is-dirty theme--light']");
+    await expect(selectedMetabolites).toHaveLength(1);
```
JellyWang-3, line 54: This is also hard coded value because I know how many check boxes it would select every time.
I had a hard time finding the corresponding xpath for verifying specific check boxes are selected every time.

e2e-test/src/tests/analyze.test.js
```diff
@@ -0,0 +1,106 @@
...
+    await expect(page).toClickXPath(vBtn('Your Datasets'));
+
+    await uploadDataset('tableTest');
+    await expect(page).toClickXPath(vListTile({ title: 'tableTest.csv ', content: 'Dataset ready for analysis.' }) + vBtn('View Data'));
+
+    // click download data
+    await expect(page).toClickXPath(vListTile({ title: 'Download Data' }));
+
+    // unselect 'Selected' for Metabolite Filter
+    await expect(page).toClickXPath("//input[@aria-checked='true'][@aria-label='Selected (0)']");
```
JellyWang-3, line 86: I'm not sure if this xpath is explicit enough

e2e-test/src/tests/analyze.test.js
dchiquit: This is perfect :+1: 
Can you add a comment documenting which checkbox should be selected? It would be very helpful if this assertion ever failed to know what metabolite should have been checked.

e2e-test/src/tests/analyze.test.js
dchiquit: Generally it is, but I don't like keying off of aria properties.
`aria` tags are special HTML tags that are used by screen readers to provide additional context about elements to people who can't see them. If you're blind and have a browser with a screen reader, the screen reader should use that extra information to read to you what your hovering over on the screen.
Try `//div[label[contains(.,'Selected (5')]]//input`

e2e-test/src/tests/analyze.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+    // verify unselected
+    await page.waitForXPath("//input[@aria-checked='false'][@aria-label='Selected (0)']");
```
dchiquit, line 89: These verifies are just checking that the select button is rendered properly as unchecked, not that the table is actually deselecting anything.
There's an element in the top left of the table: `//div[@class='column-header']/div[@class='column-header-cell']`
It shows the dimensions of the table, like `26 x 31`. Every time I select/deselect something, the dimensions change. You should be able to use that as thing to assert on.

e2e-test/src/tests/download.test.js
```diff
@@ -0,0 +1,72 @@
...
+    await page.waitForXPath("//div[@class='column-header']/div[@class='column-header-cell'][contains(text(), '15 x 17')]");
+
+    // select Transpose Table
+    await expect(page).toClickXPath("//label[text()='Transpose Table']");
+
+    //  verify it is selected
+    await page.waitForXPath("//div[@class='column-header']/div[@class='column-header-cell'][contains(text(), '17 x 15')]");
+  });
+
+  // TODO: implement code to allow download action in headless mode
```
JellyWang-3, line 46: Is this a ok comment for the download test?

dchiquit: :+1:

e2e-test/src/tests/anova.test.js
```diff
@@ -0,0 +1,51 @@
+import { vBtn, vListTile } from 'jest-puppeteer-vuetify';
+import { uploadDataset } from '../util';
+
+
+describe('download dataset', () => {
+  it('analyze table and select matabolite', async () => {
```
dchiquit, line 6:
```suggestion
  it('analyze table and select metabolite', async () => {
```

e2e-test/src/tests/download.test.js
```diff
@@ -0,0 +1,72 @@
...
+    /* eslint prefer-template: "off" */
+    await expect(page).toClickXPath("//*[text()='C26 - C26.Fol']" + vBtn({ title: 'Adds the highlighted Metabolites to the selected set' }));
+    await expect(page).toClickXPath("//*[text()='C26 - Fol']" + vBtn({ title: 'Adds the highlighted Metabolites to the selected set' }));
+    await expect(page).toClickXPath("//*[text()='C26 - Veh']" + vBtn({ title: 'Adds the highlighted Metabolites to the selected set' }));
+
+    // click download data
+    await expect(page).toClickXPath(vListTile({ title: 'Download Data' }));
+
+    // unselect 'Selected' for Metabolite Filter
+    await expect(page).toClickXPath("//div[label[contains(.,'Selected (7')]]//input");
```
dchiquit, line 28:
```suggestion
    await expect(page).toClickXPath("//div[label[contains(.,'Selected (7)')]]//input");
```

e2e-test/src/tests/download.test.js
```diff
@@ -0,0 +1,72 @@
...
+    // unselect 'C26' for Sample Filter
+    await expect(page).toClickXPath("//div[label[contains(.,'C26 (5)')]]//input");
+
+    // verify 5 rows are unselected
+    await page.waitForXPath("//div[@class='column-header']/div[@class='column-header-cell'][contains(text(), '15 x 17')]");
+
+    // select Transpose Table
+    await expect(page).toClickXPath("//label[text()='Transpose Table']");
+
+    //  verify it is selected
```
dchiquit, line 42:
```suggestion
    //  verify rows and columns are switched
```

e2e-test/src/tests/transform.test.js
```diff
@@ -0,0 +1,49 @@
+import { vBtn, vListTile } from 'jest-puppeteer-vuetify';
+import { uploadDataset } from '../util';
+
+
+describe('download dataset', () => {
```
dchiquit, line 5: Can you change the describe block for this and the ANOVA test? They both still say `download dataset`

PR #613, opened 2020-08-06
PLSDA

mvandenburgh:
Integrates the PLSDA R script into the opencpu server, adds a flask endpoint for it, and displays Score and Loadings plots in the client. Also adds tests for PLSDA function.


viime/views.py
```diff
@@ -724,6 +725,56 @@ def get_pca_overview(validated_table: ValidatedMetaboliteTable):
...
+
+    x = [[] for i in range(len(measurements))]
+
+    for i in range(num_of_components):
+        for j, score in enumerate(scores.get(f'variates.X.comp{i+1}')):
+            x[j].append(score)
+
+    explained_variances = [
+        scores.get(f'explained_variance.comp.{i+1}') for i in range(num_of_components)
+    ]
```
jbeezley, line 770: Avoid making two loops to preallocate.  Instead, use `.append`.  You might even group the three loops over `num_of_components` together into a single loop.

tests/test_analyses.py
```diff
@@ -37,3 +37,83 @@ def test_roc(client):
     for key in keys:
         for value in resp.json[key]:
             assert value in {'Inf', '-Inf'} or 0 <= value <= 1
+
+
+def test_plsda(client):
+    for i in range(2):  # run test on two datasets
```
jbeezley, line 43: You might parametrize this test rather than doing both together.  https://docs.pytest.org/en/stable/parametrize.html#parametrize-basics

web/src/components/vis/PlsdaPage/ScorePlot.vue
```diff
@@ -0,0 +1,95 @@
...
+      :row-labels="rowLabels",
+      :colors="groupLevels",
+      :group-labels="groupLabels",
+      :eigenvalues="eigenvalues",
+      :columns="columns",
+      :pc-x="pcX",
+      :pc-y="pcY",
+      :show-ellipses="showEllipses")
+  //- template(v-slot:help)
+  //-   include ../help/ScorePlotHelp.pug
```
subdavis, line 94: remove

web/src/components/vis/PlsdaPage/ScorePlot.vue
```diff
@@ -0,0 +1,95 @@
...
+    columns: {
+      type: Array,
+      required: true,
+    },
+  },
+  computed: {
+    ready() {
+      return this.$store.getters.ready(this.id);
+    },
+  },
```
subdavis, line 68: unused var.  If this is removed, the id prop will also be unused.

web/src/components/vis/PlsdaPage/ScorePlot.vue
```diff
@@ -0,0 +1,95 @@
...
+
+  methods: {
+    maybeData(key, dflt) {
+      const {
+        plot,
+      } = this;
+      return plot?.data ? plot.data[key] : dflt;
+    },
+  },
```
subdavis, line 77: Unused method

web/src/components/vis/PlsdaPage/ScorePlot.vue
```diff
@@ -0,0 +1,95 @@
...
+};
+</script>
+
+<template lang="pug">
```
subdavis, line 81: Don't use pug in new code.

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+          this.numComponents = numComponents;
+        }
+      },
+      immediate: true,
+    },
+  },
+};
+</script>
+
+<template lang="pug">
```
subdavis, line 126: Don't use pug in new code.

web/src/components/vis/PlsdaPage/LoadingsPlot.vue
```diff
@@ -0,0 +1,52 @@
...
+    loadings: {
+      type: Array,
+      required: true,
+    },
+  },
+
+};
+</script>
+
+<template lang="pug">
```
subdavis, line 43: Don't use pug in new code.

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+        return this.plot.data.scores.x;
+      }
+      return [];
+    },
+    eigenvalues() {
+      if (this.plot?.data?.scores?.sdev) {
+        return this.plot.data.scores.sdev;
+      }
+      return [];
+    },
```
subdavis, line 64: For consistency, could you use the `maybeData` method in these functions?

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+    rowLabels() {
+      if (this.dataset?.row?.data) {
+        return this.dataset.row.data.filter((row) => row.row_type === 'sample').map((row) => row.row_name);
+      }
+      return [];
+    },
+    groupLabels() {
+      if (this.dataset?.validatedGroups?.data) {
```
subdavis, line 72: In other plots, this is part of the plot response object.  Why is it different here?

Also, could you document what `validatedGroups` is in `store/index.js` ?  I don't know what it is or what it's supposed to be used for.

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+    },
+    eigenvalues() {
+      if (this.plot?.data?.scores?.sdev) {
+        return this.plot.data.scores.sdev;
+      }
+      return [];
+    },
+    rowLabels() {
+      if (this.dataset?.row?.data) {
+        return this.dataset.row.data.filter((row) => row.row_type === 'sample').map((row) => row.row_name);
```
subdavis, line 67: In other plots, this is part of the plot response object. Why is it different here?

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+    columns() {
+      if (this.dataset?.column?.data) {
+        return this.dataset.column.data;
+      }
+      return [];
+    },
+
+    groupLevels() {
+      if (this.dataset?.groupLevels) {
+        return this.dataset.groupLevels;
```
subdavis, line 89: is `groupLevels` just an array of the sample groups in a dataset?

web/src/components/vis/PlsdaPage/LoadingsPlot.vue
```diff
@@ -0,0 +1,52 @@
...
+vis-tile(title="PLS-DA Loadings Plot", svg-download)
+  loadings-plot(
+      :points="loadings",
+      :pc-x="pcX",
+      :pc-y="pcY",
+      :show-crosshairs="showCrosshairs")
+  //- template(v-slot:help)
+  //-   include ../help/LoadingsPlotHelp.pug
```
subdavis, line 51: remove

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+      immediate: true,
+    },
+  },
+};
+</script>
+
+<template lang="pug">
+vis-tile-large(title="Partial Least Squares Discriminant Analysis", :loading="plot.loading")
+  template(#controls)
+    v-toolbar.darken-3(color="primary", dark, flat, dense, :card="false")
```
subdavis, line 129: remove `:card`, default is false.

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+              hide-details,
+              type="number",
+              min="1",
+              outline,
+              :disabled="plot.loading",
+              label="Number of Components",
+              @change="changePlotArgs({ num_of_components: $event });",
+              v-model="numComponentsVal")
+
+    v-toolbar.darken-3(color="primary", dark, flat, dense, :card="false")
```
subdavis, line 144: same

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -0,0 +1,227 @@
...
+      handler(val) {
+        const numComponents = Number.parseInt(val, 10);
+        if (!Number.isNaN(numComponents)) {
+          this.numComponents = numComponents;
+        }
+      },
+      immediate: true,
+    },
+  },
+};
```
subdavis, line 123: If you find yourself setting data inside a watcher, it's often the case you should use a computed prop instead.

Now I remember that, for some reason, @dchiquit wrote these this way because of https://github.com/girder/viime/pull/531

He said that if you backspace the input box, the vis will explode because you called `parseInt` on the empty string.  However, we now have 3 duplications of this code.  1 uses the watch-set style, the other `vis/scorePlotTile.vue` uses regular computed functions.  Both work.

Hoping @dchiquit can provide some insight about why his watch-update fix was necessary in `vis/PcaPage/PcaPage.vue` but not `vis/scorePlotTile.vue`.

subdavis (dismissed):

viime/views.py
```diff
@@ -724,6 +725,56 @@ def get_pca_overview(validated_table: ValidatedMetaboliteTable):
...
+    max_components = min(rows, cols)
+    if num_of_components is None:
+        num_of_components = max_components
+    if num_of_components > max_components:
+        return jsonify({
+            'error': 'invalid num_of_components'
+        }), 400
+
+    scores = plsda(measurements, groups, num_of_components, 'scores')
+    loadings = plsda(measurements, groups, num_of_components, 'loadings')
```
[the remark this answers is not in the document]
mvandenburgh: I looked into this, but the scores and loadings don't fit into the same R `dataframe` since they are different sizes. The PCA analysis also uses two opencpu requests to retrieve its scores and loadings, but in that case there is two separate flask endpoints -
`GET /csv/<uuid:csv_id>/plot/pca`
`GET /csv/<uuid:csv_id>/plot/loadings`

I can split the PLS-DA scores and loadings into two endpoints like that, but I believe they will always be called at the same time anyway so performance will be the same. What do you think?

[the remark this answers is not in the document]
jbeezley: I see.  I think there are ways of returning more unstructured data types, but I never instrumented it.  This is fine for now.

viime/views.py
mvandenburgh: Is this what you mean by using `.append`? https://github.com/girder/viime/pull/613/commits/ed4451db8b56eb083bfcedbbfb42a1a95de24889

web/src/components/vis/PlsdaPage/PlsdaPage.vue
mvandenburgh: Yes, I believe so.

web/src/components/vis/PlsdaPage/PlsdaPage.vue
mvandenburgh: It was different because the PLS-DA endpoint didn't send back the `rows` and `labels` like the PCA one. I just changed it to do that so it's more consistent. I'll look at the `validatedGroups` and document that

web/src/components/vis/PlsdaPage/PlsdaPage.vue
mvandenburgh: See https://github.com/girder/viime/pull/613#discussion_r472363905

web/src/components/vis/PlsdaPage/PlsdaPage.vue
mvandenburgh: Done. I had to modify the `maybeData` function to allow nested objects.

web/src/components/vis/PlsdaPage/PlsdaPage.vue
jeffbaumes: @dchiquit could you provide some input so we can get this closer to merging?

subdavis:
Can't merge this yet.

https://github.com/girder/viime/pull/613/files#diff-5044477b413042739d1da31a438bb261R312

The sdev , which is used for eigenvalues, is coming back as [[v1, v1, v1, v1, v1, v1, ...], [v2, v2, v2 ,v2, v2...], ...., [vn, vn, vn...]  instead of [v1, v2, v3, ... vn]

I think it's probably a simple mistake, but I can't really make sense of the R code.  I tried removing the as.list and got this error

```
arguments imply differing number of rows: 77, 63

In call:
data.frame(variates = mod_plsda$variates[1], explained_variance = mod_plsda$explained_variance$X)
```

Where I come from, casting things to a list shouldn't change the length of the iterable.

jeffbaumes:
@subdavis I believe the issue is getting all the needed data back from R in a single dataframe. The `scores` mode returns both `x` (num_samples x num_components) and `sdev` (1 x num_components) which are of different sizes. The way it seems things were worked around was to return one dataframe size (num samples x [2 x num_components]) where the `sdev` columns repeat the same value. Getting the first value out of each when converting for delivery over the REST endpoint seems to get it in the right form for the client.

Another option would be to have three modes (and three calls to OpenCPU): 'scores', 'loadings', and 'sdev' so that we don't need to do this weirdness, but the current approach seems reasonable.

subdavis:
@jeffbaumes thanks for the fix. I'm curious if you think that's the "correct" solution or it's just covering a problem somewhere in the R script?  I think it'd be best to avoid transferring all the additional data transfer between opencpu and python if possible, but if the R script is producing a legitimate PLSDA analysis and we simply need to filter it for our visualization, that's fine.

Should the nested array be happening?

jbeezley:
Right now, we can only return data frames (basically a matrix).  It is possible to return other data types such as a heterogeneous list, but that would require some additional logic in the code.  I think the latency of communicating with opencpu dominates the time taken, so duplicating the data is likely not significant here.

PR #614, opened 2020-08-07
Convert PcaPage to use Typescript

dchiquit:
Also convert the three components on the PcaPage, ScorePlot, ScreePlot,
and LoadingsPlot.

Using the composition API on PcaPage created a circular dependency:
```
PcaPage.js -> usePlotData.ts -> store/index.js -> analyses.js ->
PcaPage.js
```

The problem was that `analyses.js` provides a list of all available
analyses and all data required to describe them, including the
components that should be used to render them. `store/index.js` imported
`analyses.js` to dynamically determine what analyses needed plot data
stores, which had the side affect of importing all components used in
those analyses.

The solution was to remove the data required for the store from
`analyses.js` and move it to `store/index.js`. The downside is that
analysis data is now defined in two places, but it seems appropriate
that visualization info is kept in `vis/analyses.js` while store info is
kept in the store.


web/src/components/vis/PcaPage/PcaPage.vue
```diff
@@ -76,6 +76,89 @@ export default {
     },
   },
 };
+</script>-->
```
subdavis, line 79: Remove?

web/src/components/vis/PcaPage/PcaPage.vue
```diff
@@ -76,6 +76,89 @@ export default {
...
+    const numComponentsVal = ref('10');
+    const pcX = ref(1);
+    const pcY = ref(2);
+    const numComponents = ref(10);
+    const showEllipses = ref(true);
+    const showCrosshairs = ref(true);
+    const showCutoffs = ref(true);
+    const showScree = ref(true);
+    const showScore = ref(true);
+    const showLoadings = ref(true);
```
subdavis, line 117: I think this is a situation where I would have considered `reactive` instead of `refs` for brevity in the return function.

```typescript
    const plot  = reactive({
      pcXval: 1,
      pcYval: 2,
      numComponentsVal: 10,
      /* ... */
    });
    /* .... */
    return {
      ...toRefs(plot),
      ready,
    };
```

web/src/components/vis/PcaPage/PcaPage.vue
```diff
@@ -76,6 +76,89 @@ export default {
...
+      showEllipses,
+      showCrosshairs,
+      showCutoffs,
+      showScree,
+      showScore,
+      showLoadings,
+      ready,
+    };
+  },
+});
```
subdavis, line 161: What made you decide to write this with composition api rather than just object api?

web/src/components/vis/PcaPage/PcaPage.vue
dchiquit: I'd never heard of `toRefs`, that is the better solution.

web/src/components/vis/PcaPage/PcaPage.vue
dchiquit: I've been doing these very mechanically and composition API is a box on my checklist. This is actually the first component that hasn't required a mixin/use function, so it's the first time it's come up.

My gut instinct is to use composition API everywhere just to avoid having to think about two different ways to write components. In terms of aesthetics, I don't really care. I think they take up exactly the same number of lines, and this component is small enough to fit on one page so it's not hard to read in it's entirety either way.

In terms of maintainability, I feel like using composition API will make it easier to compare and copy/paste between components. I have no evidence whatsoever to support that claim, though.

I remember you saying that the object API wasn't going anywhere and that composition API had it's cons. Is there anything you can think of that should be considered here?

subdavis: I'm also on the fence.

* I have heard from others and personally feel that Object API is more readable.  I do think it provides more guardrails for newer Vue developers through its opinionated style.
* It's also the style with the most documentation support, Stack Overflow questions, etc.
* Likewise, Composition API requires more discipline to keep code sane.  It turns the general wisdom about clousures/scope pollution totally on its head, so when you use something from the parent closure, you sort of have to ask yourself "would I do this in Object API?"

* However, I agree that that the ability to bring in composition functions without another style rewrite is also valuable.
* Since it's his last day, I'd be really interested to hear @mvandenburgh's thoughts.  How does this code look to you?  Do you like it more or less than Object API?  Do you think your Vue learning process would have been different if everything were written like this?

I think there's a fair chance we'll want to use a composition function here in the future, so I think we should keep this file as-is.  But "when should we use which style" is bound to keep coming up, so it would be good if our team was on the same page about the merits of each.

waxlamp: I'm not mvandenburgh but I thought I'd throw in my pennies nonetheless.

I think it would be best to get behind one way of doing things, and thus eliminate the question of which style to use for each component. Composition API brings many positives, among which I consider *more explicitness* and *typescript support* to be the main ones.

What are the drawbacks? And I am also curious to know more about the general wisdom around "closures/scope pollution".

mvandenburgh: > Do you think your Vue learning process would have been different if everything were written like this?

I think the more opinionated style of the Object API is better suited to learning Vue. Having the different parts of a component separated into logical pieces is easier to work with and less intimidating for someone learning it for the first time, I think. I don't think this is a huge advantage for using Object vs Composition though, the larger issue is that the Object API has more documentation support and online resources as you mentioned, which was really helpful for me when I was starting off with Vue. If the Composition API had the same level of documentation/resources as the Object API I think the learning process would have been roughly the same, maybe _slightly_ more difficult because of the first issue I mentioned.

> How does this code look to you? Do you like it more or less than Object API?

I think the Composition API allows for cleaner/more readable code if you're already familiar with how Vue works. For example, I think being able to explicitly declare reactive variables like at https://github.com/girder/viime/blob/ts-pca-page/web/src/components/vis/PcaPage/PcaPage.vue#L106 makes it much easier to figure out what's going on in the code. The better support for TypeScript and the ease of using composition functions is also a big advantage for Composition API.

I haven't done anything with the Composition API outside of the one component that I converted, but based on that and all the TypeScript conversion PRs I've looked at on here, I think I prefer it over the Object API. Also, [one of the stated goals in the RFC for the Composition API is to allow for less complex code in large-scale projects](https://composition-api.vuejs.org/#motivation).

subdavis: Thanks for your feedback, everyone.  I would hedge slightly and propose the following:

1. Large, complex components are written in Composition API
1. If you suspect a composition function could add value in the near future, Composition API should be used.  That will mostly come from a spider-sense that some significant bit of state or functionality is likely to be refactored and shared.
1. Every application is organized as a tree of components, and nodes toward the top of the tree (App.vue, all views if router is being used, maybe some of their direct children) will typically use Composition API (mostly because of 1 and 2)
1. Finally, very simple components, presentation-level leaf nodes, and stateless components that may only have a few props and computed properties, should consider using Object API.   This will hopefully encourage good separation of responsibilities, keeping presentation layer logic and business logic apart.

Is that an accurate summary, @waxlamp @dchiquit @mvandenburgh or would you prefer one consistent style regardless?

> And I am also curious to know more about the general wisdom around "closures/scope pollution".

I only mean that, in a composition function, anything declared in `setup()` is available to everything else.  We had a thread last week where we talked about declaring dependencies at the top of computed functions because you can destructure them from `this`;  In composition functions, everything is already in scope.  You can't even `CTRL+F` to find the `this` keyword to see what all the references to instance variables are in a method or computed prop.

Another example: in Object API, you have separate functions for the created and mounted hooks.  Variables declared in that func get garbage-collected when it exits and they cant be used elsewhere.  

If you happen to have a complex created hook in Composition api, all your temporary vars are now available to all your computed functions and methods, and you might be tempted to use them.   You can mitigate this by being discliplined about doing all your refs at the top of setup, then computed, then watchers, then functions, and finally created hook logic.  But something you have to "remember" to do in Composition API is a code quality perk you get for free in Object.

waxlamp: > Thanks for your feedback, everyone. I would hedge slightly and propose the following:
> 
> 1. Large, complex components are written in Composition API

:heavy_check_mark: 

> 2. If you suspect a composition function could add value in the near future, Composition API should be used.  That will mostly come from a spider-sense that some significant bit of state or functionality is likely to be refactored and shared.

This is describing a situation that almost always goes from "I think it's simple now" to "oops, it wasn't simple because humans are bad at predicting the future" in a ratchet-like fashion. That's part of why I prefer to just be bold and say "always use Composition API unless there's a reason not to".

However, my position is weak until I hear about the drawbacks of using Composition API compared to Object API. Is it just the closure pollution issue? The documentation issue is somewhat of a problem, though I believe it will be less so with each passing day.

> 3. Every application is organized as a tree of components, and nodes toward the top of the tree (App.vue, all views if router is being used, maybe some of their direct children) will typically use Composition API (mostly because of 1 and 2)

:heavy_check_mark: 

> 4. Finally, very simple components, presentation-level leaf nodes, and stateless components that may only have a few props and computed properties, should consider using Object API.   This will hopefully encourage good separation of responsibilities, keeping presentation layer logic and business logic apart.
> 

I like the use of "consider" in this point of guidance. However, I'm not sure that keeping two different programming models always in mind is the best way to achieve a separation between presentation and business logic. There will likely be times that even a simple presentational component could benefit from using Composition API (even if it's nothing more than solid TypeScript support, for instance), and in general the ability to recompose bits of (pure) presentational logic in other components could be a big boon.

> Is that an accurate summary, @waxlamp @dchiquit @mvandenburgh or would you prefer one consistent style regardless?
> 

It's probably clear by now that I'm arguing for one consistent style: i.e., use Composition API unless there's an overriding reason not to. (One such reason could be that the component is so simple and so un-reusable that Object API is preferred.)

> > And I am also curious to know more about the general wisdom around "closures/scope pollution".
> 
> I only mean that, in a composition function, anything declared in `setup()` is available to everything else. We had a thread last week where we talked about declaring dependencies at the top of computed functions because you can destructure them from `this`; In composition functions, everything is already in scope. You can't even `CTRL+F` to find the `this` keyword to see what all the references to instance variables are in a method or computed prop.
> 
> Another example: in Object API, you have separate functions for the created and mounted hooks. Variables declared in that func get garbage-collected when it exits and they cant be used elsewhere.
> 
> If you happen to have a complex created hook in Composition api, all your temporary vars are now available to all your computed functions and methods, and you might be tempted to use them. You can mitigate this by being discliplined about doing all your refs at the top of setup, then computed, then watchers, then functions, and finally created hook logic. But something you have to "remember" to do in Composition API is a code quality perk you get for free in Object.

This is a good point. There are ways around it, though they may not be super palatable. The main method that comes to mind is to declare computed functions outside the component definition, and supply them with arguments that they will use internally. Then, call these functions within the `setup()` method, supplying only the refs that are needed. This could apply to the logic of the created/mounted hook as well.

However, that doesn't sound like a great way to write all our components. I can offer a much weaker defense as well: that it's easy to write bad code no matter the programming model, and part of why we get paid to do this is that we do have the discipline and care to do it right. Plus there's code review. But this is a very Guido van Rossum "we're all adults" type of argument, something I've rejected when it comes to defenses of questionable/dangerous features of Python. Still, for me this is a matter of cost and benefit, and the benefits I've gotten from using Composition API have been a bonanza so far.

jeffbaumes: I tend to agree that we should prefer a single style, but also would not be horrified if we ended up leaving some object components in here. To help people re-cast object code as composition code (say from stackoverflow answers) we could point people to blogs like [this](https://vuedose.tips/easily-switch-to-composition-api-in-vuejs-3/). I think for the most part it would be mechanical translation. It's sort of like what we had been doing with Pug (i.e. we couldn't just copy-paste solutions from Vuetify examples or stack overflow). But in this case I believe composition is more forward-looking than going off the mainstream like we were doing with Pug. Plus copy/pasting html template code to set up a certain Vuetify look is probably much more common than wanting to directly copy/paste props/watchers/etc. in JavaScript.

dchiquit: The design constraints of the Composition API seem like they apply to more complex components. If a component is heavy enough to have created/mounted hooks and enough variables to be confusing, it's large enough to warrant using Composition API for purely organizational reasons. It's not flawless, but no API is. I think it would be better to adopt Composition API universally and start familiarizing ourselves with the gotchas we need to be aware of.

web/src/components/vis/PcaPage/PcaPage.vue
```diff
@@ -123,136 +46,225 @@ export default defineComponent({
       return pcaReady && loadingsReady;
     });
 
-    watch(pcXval, (newPcXval) => {
+    watch(() => plot.pcXval, (newPcXval) => {
```
subdavis, line 49: Why is this function necessary?

dchiquit: I think I was being defensive. Documentation suggests that `reactive` creates deep copies, so presumably `plot.pcXval` is a `ref`, but maybe it wasn't.

subdavis: I'm not sure I follow, sorry.  `reactive` should indeed create deep copies during initialization, but I don't understand how that relates to this wrapper.  `plot.pcXVal` is a reactive proxy object, but it is not a `Ref` type.

Issue #622, opened 2020-08-12
Including example datasets in local installation


joannawolthuis:
Hi crew! Thanks for the updated instructions :) I can properly load up the server now! I was wondering if examples are included in the local installation ? For me I am forever stuck here:
![image](https://user-images.githubusercontent.com/16805477/90006746-ca475180-dc99-11ea-93c1-53834e912d07.png)
So either I'd have a 'no examples available' prompt here, or include them (which would be nice for further testing)

waxlamp:
Hi @joannawolthuis!

I investigated a bit, and for me I always see a panel that says "no datasets found"--*unless* the server is not running, in which case I see the situation you have described (I opened #625 to address this, putting up an error message instead of just forever spinning).

Can you confirm that your server is running? A `docker ps` command should show two containers named `docker_backend_1` and `docker_opencpu_1`. If you do have those running, we'll have to dig a bit deeper.

jeffbaumes:
Regardless of if we can get this working, add a pointer to an example CSV that works with the system from the README.

waxlamp:
Hi @joannawolthuis!

After merging #649, there are now sample datasets included in the repository, along with instructions on how to use them to set things up in a local build. Please let me know if this is working for you (reopen this issue if there are still problems). Thanks!

PR #625, opened 2020-08-12
Add specific error message to sample dataset panel if server is down

waxlamp:
This PR prevents the "loading" icon from continuing to spin in the case where a server is not running (or otherwise not reachable) at the time the sample datasets panel appears. In that case, the endpoint for getting the sample datasets return a 500, but the code was not equipped to handle this error condition.

Addresses in part #622.

### DIFF
--- a/web/src/components/SampleBrowser.vue
+++ b/web/src/components/SampleBrowser.vue
@@ -7,6 +7,7 @@ export default {
     return {
       loading: true,
       samples: [],
+      error: false,
     };
   },
   computed: {
@@ -14,7 +15,15 @@ export default {
   },
   async created() {
     this.loading = true;
-    this.samples = (await SampleService.list()).data;
+
+    try {
+      const samples = await SampleService.list();
+      this.samples = samples.data;
+    } catch (err) {
+      this.samples = [];
+      this.error = true;
+    }
+
     this.loading = false;
   },
   methods: {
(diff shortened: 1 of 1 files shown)


web/src/components/SampleBrowser.vue
```diff
@@ -7,14 +7,23 @@ export default {
...
   computed: {
 
   },
   async created() {
     this.loading = true;
-    this.samples = (await SampleService.list()).data;
+
+    try {
+      const samples = await SampleService.list();
+      this.samples = samples.data;
```
dchiquit, line 21:
```suggestion
      const { data } = await SampleService.list();
      this.samples = data;
```

web/src/components/SampleBrowser.vue
```diff
@@ -7,14 +7,23 @@ export default {
...
   },
   async created() {
     this.loading = true;
-    this.samples = (await SampleService.list()).data;
+
+    try {
+      const { data } = await SampleService.list();
+      this.samples = data;
+    } catch (err) {
+      this.samples = [];
```
subdavis, line 23: superfluous, I think.  `this.samples` can only ever be `[]` at the time this line runs.

waxlamp:
```suggestion
```

PR #629, opened 2020-08-13
Add test for factor analysis

mvandenburgh:
Adds a test for the factor analysis endpoint/opencpu function. (Forgot to do this when I originally implemented the ROC stuff)

### DIFF
--- a/tests/test_analyses.py
+++ b/tests/test_analyses.py
@@ -1,36 +1,10 @@
-from io import BytesIO
 import json
-from pathlib import Path
 
 from flask import url_for
 
 
-def upload_test_dataset(client, file, changes=None):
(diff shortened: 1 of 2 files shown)


tests/test_analyses.py
```diff
@@ -5,25 +5,32 @@
 from flask import url_for
 
 
-def test_roc(client):
-    with (Path(__file__).parent / 'roc.csv').open('rb') as f:
+def upload_test_dataset(client, file, changes=None):
```
mvandenburgh, line 8: Refactor test dataset uploading into a separate function

jbeezley: This would be cleaner as a parametrized fixture.  https://docs.pytest.org/en/stable/fixture.html#parametrizing-fixtures

PR #641, opened 2020-08-19
Disable zoom controls on all vis tiles

### LINKED ISSUE #616: Zooming in/out resizes the window but doesn't resize the plots itself
via @jkanche

waxlamp:
This removes the zoom controls from all vis tiles, to address the fact that most of our plots did not attach any semantics to these controls, leading to bad UX.

Fixes #616

### DIFF
--- a/web/src/components/vis/VisTile.vue
+++ b/web/src/components/vis/VisTile.vue
@@ -75,13 +75,6 @@ v-flex.white.rounded.main(shrink=1, :class="scaleClass")
         slot(name="help")
     v-spacer
     v-toolbar-items
-      v-btn(@click="setScaleIndex(scaleIndex - 1)", :disabled="scaleIndex === 0", icon)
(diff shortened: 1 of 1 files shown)


jeffbaumes:
Modify to fully take out zoom control options to resize plots.

subdavis:
This is how it looks for me now.  Is this intended?  I think the plots may be a bit too small.

Screenshot from 2020-08-24 09-27-31

jeffbaumes:
I think the plot size should be made quite a bit larger, those are small. And obviously the box plot should not be cut off.

waxlamp:
I've restored the default plot size now (via 06666b3). Please take another look and let me know if it looks ok now.

As for cutting off the box plot--we seem to be rendering it in just enough horizontal space to accommodate the rightmost outlier, and it scrolls vertically by default in order to maintain a readable font size.

Issue #638, opened 2020-08-19
Add an infobox to the correlation network view to explain how it works


waxlamp:
In particular:
- [ ] meaning of edge thickness
- [ ] meaning of edge color

See https://github.com/girder/viime/issues/621.

PR #644, opened 2020-08-21
Add a popover help panel to VisTileLarge, and enable it in all analysis pages

waxlamp:
This PR adds a help button to each analysis page which activates a popover that shows the same analysis description that is shown in the master analysis display.

This partially addresses #638 (to completely fix that issue, we need to clarify the help text for the correlation network as well).

### DIFF
--- a/web/src/components/vis/analyses.js
+++ b/web/src/components/vis/analyses.js
@@ -11,7 +11,7 @@ import { plot_types } from '../../utils/constants';
 import { correlation_methods } from './constants';
 import vuetify from '../../utils/vuetifyConfig';
 
-export default [
+const analysisList = [
   {
     path: 'pcapage',
     name: 'Principal Component Analysis',
@@ -164,3 +164,17 @@ export default [
     icon: vuetify.icons.pca,
   },
 ];
+
+const analysisTable = (() => {
+  let table = {};
+  analysisList.forEach((entry) => {
+    table[entry.path] = entry;
+  });
+
+  return table;
+})();
+
+export default analysisList;
+export {
+  analysisTable,
+};
--- a/web/src/components/vis/analyses.js
+++ b/web/src/components/vis/analyses.js
@@ -166,7 +166,7 @@ const analysisList = [
 ];
 
 const analysisTable = (() => {
-  let table = {};
+  const table = {};
   analysisList.forEach((entry) => {
     table[entry.path] = entry;
   });
--- a/web/src/components/vis/analyses.js
+++ b/web/src/components/vis/analyses.js
@@ -165,16 +165,13 @@ const analysisList = [
   },
 ];
 
-const analysisTable = (() => {
-  const table = {};
-  analysisList.forEach((entry) => {
-    table[entry.path] = entry;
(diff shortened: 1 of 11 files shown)


web/src/components/vis/analyses.js
```diff
@@ -164,3 +164,17 @@ export default [
...
 ];
+
+const analysisTable = (() => {
+  const table = {};
+  analysisList.forEach((entry) => {
+    table[entry.path] = entry;
+  });
+
+  return table;
+})();
```
subdavis, line 175: Could we...

1. call this `analysisMap` instead, because `table` and `list` are really the same thing to me?
2. use `Object.freeze()` before the return to prevent confusion about how this new thing should be used? 
3. use reduce instead of the IIFE?

**Note: I wrote this in github, it's untested**

```suggestion js
const analysisMap = Object.freeze(analysisList.reduce((map, entry) {
    map[entry.path] = entry;
    return map;
}, {}));
```

waxlamp: Excellent recommendations. See 86f2e12 -- I did things slightly differently than you suggested.

web/src/components/vis/analyses.js
```diff
@@ -164,3 +164,14 @@ export default [
     icon: vuetify.icons.pca,
   },
 ];
+
+// Create a path-indexable map version of the analysis list.
+const analysisMap = Object.freeze(analysisList.reduce((map, entry) => ({
+  ...map,
+  [entry.path]: entry,
+}), {}));
```
subdavis, line 172: @waxlamp while subtle, this results in something rather different than what I was suggesting. 

* a new object is created at every iteration of the loop.  This is unnecessary, since reduce() takes an initializer as its second param and you can use the same object for the whole loop.  This is at worst an `O(N)` cost
* somewhat more tricky, though, is that spread.  You can never really know what babel will do with language features like this, but in this case, I think it gets transpiled into an `Object.keys()` call.  What would have been a `O(n)` loop is now `O(n*2 * m * c)` where N is array length, M is keys per object, and C is some constant because it looks like the code babel generates does a bunch of looping and filtering.

I wouldn't have bothered to mention this, but I'd regret if this code to get copied somewhere else and used under different circumstances.

[Here's the worst-case scenario](https://babeljs.io/repl#?browsers=defaults%2C%20not%20ie%2011%2C%20not%20ie_mob%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYewdgzgLgBAhjAvDA3gXwNwChSVgIyVQDpS5Mg&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=env%2Creact%2Cstage-2%2Cenv&prettier=false&targets=&version=7.11.4&externalPlugins=)

waxlamp: Thanks for pointing this out (and for not explicitly calling me out for trying to be too clever :stuck_out_tongue:). I'll open a PR to fix this.

PR #646, opened 2020-08-24
Strip out columns with zero variance in factor analysis

### LINKED ISSUE #628: Factor analysis sentry error
Got this error from the factor analysis endpoint on the group prediction page here https://viime.org/#/app/pretreatment/dbe03b9d-8007-4508-9de2-a138d7402580/analyze/roc

```
cannot rescale a constant/zero column to unit variance

In call:
prcomp.default(m.df, center = T, scale = T)
```

https://sentry.io/organizations/kitware-data/issues/?project=1814175

jbeezley:
The issue in #628 was caused by several columns that are all zero.  The principal component analysis is set up to rescale all columns to zero variance, which it clearly can't do in this case.  This removes all of those columns before executing the rest of the code.  I'm not sure if this is the correct thing to do, but it does fix the issue.  It might be worth asking Lillian.

Fixes #628

### DIFF
--- a/devops/viime/R/analyses.R
+++ b/devops/viime/R/analyses.R
@@ -252,6 +252,7 @@ factor_analysis <- function(measurements, threshold) {
   #- (Eigenvalues higher than 1) -#
 
   # Principal Component Analysis to get eigenvalues
+  m.df <- m.df[ , which(apply(m.df, 2, var) != 0)]
   pca_a <- prcomp(m.df, center=T, scale=T)
 
   #getting the eigenvalues


jeffbaumes:
I'm curious: did the low-variance columns show up as warnings at ingestion time? Instead of silently removing columns from analysis, I believe it would be better to do either of the following:

1. Turn low-variance warnings into errors so they need to be removed before proceeding to analysis. This may be easiest.
2. Show a descriptive warning on the factor analysis page if the analysis fails for this reason. You could check for any no-variance columns in the factor analysis and bail out early from R. Would there be a way to signal an error message in the return value (a special kind of dataframe maybe?) and pass descriptive R edge case errors through to the client? May not be worth going down this trail.

jbeezley:
> I'm curious: did the low-variance columns show up as warnings at ingestion time?

Yes.

I think I would add a new case for *exactly* zero variance and make that an error.  Perhaps even better, we could mask zero variance columns at ingestion time.  The current [ingest code](https://github.com/girder/viime/blob/master/viime/models.py#L62) was never really intended to be so simple.

jeffbaumes:
> I think I would add a new case for exactly zero variance and make that an error. Perhaps even better, we could mask zero variance columns at ingestion time. The current ingest code was never really intended to be so simple.

I think it is less surprising for the user to see the columns with errors and need to hide them, rather than having columns silently disappear on ingest. They may wonder, "Where's that metabolite?" in downstream analyses. So +1 from me to just make it an error.

Also, is there another edge case where a column has a constant value plus some missing values (or all missing values), and depending on the missing value estimation technique it might end up zero variance (or not)? Ideally through whatever means we can enforce that before going to analysis there are no zero-variance columns.

waxlamp:
Just FYI, here is what Lilian had to say about the solution in this PR:

> Yes, that is a good way to solve it. I think the codes I share for factor analysis and PLS are assuming the treatment of the data is happening before and we wouldn’t have the problem of zeros. I didn’t want to use the center and scaling functions directly on the factor analysis since people could want to try different transformations instead of just centering and scaling.
>
> I think it is better to take out the columns with zero variance as Jonathan is saying.
>
> Let me know if that makes sense or if we want to change that.

jbeezley:
> I think it is less surprising for the user to see the columns with errors and need to hide them, rather than having columns get silently hidden on ingest. They may wonder, "Where's that metabolite?" in downstream analyses. So +1 from me to just make it an error.

I played around with the idea in #652, but I can close it if you think the current behavior is better.

jeffbaumes:
I like #652. I wasn't thinking of all the other smart things we could do here, other than zero-variance masking. We may just want a better indicator of how many rows/columns are hidden so that the user can quickly see what "smart" decisions happened on ingest.

PR #657, opened 2020-09-01
Use toRef instead of computed values

dchiquit:
There are some cases where a prop must be passed as a reactive value to
a child component. Since individual props aren't reactive, the initial
solution was to wrap them in a computed property. I have since been
pointed to `toRef`, which solves this exact problem.


subdavis (approved):
sanctioned by the official docs :+1:   https://composition-api.vuejs.org/api.html#toref

LGTM

PR #658, opened 2020-09-02
TS toolbar mixins

dchiquit:
Convert the toolbar mixins to TypeScript use files and convert all components which use them.
The components could stand to be cleaned up a bit, but the conversion was messy enough.


web/src/components/toolbar/use/useMetabolite.ts
```diff
@@ -0,0 +1,155 @@
+import {
+  PropType, computed, watch, Ref, ref,
+} from '@vue/composition-api';
+import { colors } from '../../../utils/constants';
+
+// TODO this is a subset of a type from Vuex
```
dchiquit, line 6: I think it might be time for a `types.ts` file that contains common types like this that are referenced frequently. Thoughts?

subdavis: Fine with me.

subdavis (approved):
:100:

PR #661, opened 2020-09-04
Convert util mixins to TS

dchiquit:
The util mixins have been deleted and moved into the components which used them.

This adds a circular dependency between `routes/index.js` and
`Pretreatment.vue`. `Pretreatment` is required in `routes` because it is
navigable in one of the default routes, while `routes` is required in
`Pretreatment` because it needs to trigger navigation to different
components. This was not a problem because `$routes` was injected into
2.0 component definitions, but this is no longer possible with
composition API.


web/src/components/Pretreatment.vue
```diff
@@ -1,160 +1,269 @@
...
-    valid(dataset) {
-      return this.$store.getters.valid(dataset.id);
-    },
-    isMerged(dataset) {
-      return this.$store.getters.isMerged(dataset.id);
-    },
-    problemNav(problem, id) {
+  setup(props) {
+    // TODO Use injected router instead of importing to avoid circular import
+    const router = inject('router') as VueRouter;
```
subdavis, line 35: I recall that you asked recently how to get this inside a composition function: import or inject.  I don't remember what I said, but I went to look at the vue router next docs and inject will be preferred, but you'll be able to use builtin composition functions from the plugin like this:

https://next.router.vuejs.org/api/#composition-api

You may have already known that.  Anyway, this is fine with me.

web/src/components/Pretreatment.vue
```diff
@@ -1,160 +1,269 @@
...
+    fill-height="fill-height"
+  >
+    <v-navigation-drawer
+      class="navigation"
+      floating="floating"
+      permanent="permanent"
+      style="min-width: 220px; width: 220px;"
+      touchless="touchless"
+      disable-resize-watcher="disable-resize-watcher"
+      stateless="stateless"
```
subdavis, line 91: Looks like there are still some boolean tags here.

e2e-test/src/tests/cleanUp.test.js
```diff
@@ -19,16 +19,19 @@ describe('relabel dataset', () => {
...
     //  verify the dataset is failing
+    await new Promise((resolve) => setTimeout(resolve, 1000));
     await page.waitForXPath(vListTile({ title: 'cleanUpTest.csv' }) + vIcon({ cssClass: 'mdi mdi-alert theme--light warning--text' }));
 
     // select the third column(column 'C') and mark it as a 'Group'
     await expect(page).toClickXPath("//div[contains(@class,'column-header-cell type-measurement')][contains(text(), 'C')]");
     await expect(page).toClickXPath("//input[@value='group']");
 
     //  verify the dataset is ready to process
+    await new Promise((resolve) => setTimeout(resolve, 1000));
```
subdavis, line 34: why was it necessary to put this wait before the next await?  Shouldn't the `waitForXPath` make this unnecessary?

This seems confusing to me, but I just may not be familiar with this testing framework.

subdavis (approved):
LGTM.

e2e-test/src/tests/cleanUp.test.js
dchiquit: I put these wherever a test run failed to find something that was on the page. `waitForXPath` uses the puppeteer `waitForXPath`, which I cannot guarantee the reliability of. The tests test reliably with them there and not when they are absent, so while constant waits are an antipattern, I personally think that sometimes they are worth it to amortize debugging time.

subdavis: > they are worth it to amortize debugging time.

I've never really looked at it that way, but I think that's a really valid concession to make.  

:+1:

PR #667, opened 2020-09-11
Add server support for oplsda analysis

jbeezley:
This returns data that should be compatible with the plsda endpoint.  The only thing I'm unsure about is the "explained variance" or "sdev" that gets returned with the scores.  Lilian's comment in the example code states that `R2Y` is "R2 for main axis", so my assumption is that "sdev" should be the square root of that value, but it should be verified.


dchiquit (approved):
LGTM, assuming the R^2 value is correct.

jbeezley:
Not sure about R^2 yet, but we can change it later if needed.

PR #668, opened 2020-09-16
Ts plsda

dchiquit:
Convert PlsdaPage to TypeScript.

Also, I noticed that changing Number of Components doesn't do anything. It is changing the plot argument, but was not triggering a page reload. I hacked something in to invalidate the plot and force a reload, but I'm happy to back that out if it's too dirty.

### DIFF
--- a/web/src/components/vis/PlsdaPage/PlsdaPage.vue
+++ b/web/src/components/vis/PlsdaPage/PlsdaPage.vue
@@ -1,29 +1,29 @@
-<script>
+<script lang="ts">
 import VisTileLarge from '@/components/vis/VisTileLarge.vue';
 import LayoutGrid from '@/components/LayoutGrid.vue';
 import ScorePlot from './ScorePlot.vue';
 import LoadingsPlot from './LoadingsPlot.vue';
 import plotData from '../mixins/plotData';
+import { computed, defineComponent, reactive, ref, toRef, toRefs, watchEffect } from '@vue/composition-api';
+import usePlotData from '../use/usePlotData';
+import store from '../../../store';
 
-export default {
-  components: {
-    ScorePlot,
-    LoadingsPlot,
-    VisTileLarge,
-    LayoutGrid,
-  },
-
-  mixins: [plotData('plsda')],
-
+export default defineComponent({
   props: {
     id: {
       type: String,
       required: true,
     },
   },
-
-  data() {
-    return {
+  components: {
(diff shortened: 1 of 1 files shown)


waxlamp:
Isn't the number of components control meant for the scree plot?

dchiquit:
PLSDA only has a score and a loadings plot, no scree plot. Neither of those plots take a `numComponents`. The R analysis does take a `num_components` argument, but changing it in the UI does not trigger a refresh, as you can see in the network tab of the inspector.

web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -35,85 +36,55 @@ export default {
...
+      const numComponents = Number.parseInt(controls.numComponentsVal, 10);
+      if (!Number.isNaN(numComponents)) {
+        controls.numComponents = numComponents;
+      }
+    });
+
+    return {
+      plot,
+      changePlotArgs,
+      ...toRefs(controls),
```
subdavis, line 76: as a matter of style, if a group of reactive properties are coupled such that you think they all belong within the same `const controls = reactive({ ... })`, then it would probably make sense to `return { controls };`

Then you can use  `{{ controls.whatever }}` in the template. 

I consider `toRefs()` to be a tool for composition functions and not really something you'll use in SFCs.  

Again, this is more of a style thing.  I'm not sure if @dchiquit chose `toRefs` intentionally or thought it was required for all `reactive()` objects.

subdavis (approved):

web/src/components/vis/PlsdaPage/PlsdaPage.vue
subdavis: You can also sorta take this with a grain of salt.  Spread (`...`) is my new pet peeve, mostly because it's hard to know/understand how babel with transform it.  I've become increasingly comfortable with spread over the last year and I've caught myself using it in nested loops and other dumb places.  I'm sorta hedging towards "we should use fewer babel features". 

Anyway, I'm projecting here is my point.

dchiquit: I was mostly trying to minimize footprint and save verbosity, but now that I consider it properly `v-model="controls.showEllipses"` is more legible to me than `v-model="showEllipses"`. I will change that.

subdavis (approved):

PR #669, opened 2020-09-17
Oplsda

dchiquit:


subdavis (approved):
This looks exactly like I would expect, it's mostly just a variation on PLSDA.

I haven't pulled an run this, I won't have time until probably the end of next week, but the code looks great and IMO you should merge it.

PR #673, opened 2020-09-24
Plsda improvements

dchiquit:
Backend changes:
* Add VIP scores to PLSDA and OPLSDA in analysis.R
* Add num_of_components to OPLSDA to match PLSDA
* Add Q^2 values to OPLSDA
* Add R^2 and Q^2 values to PLSDA to match OPLSDA
* Pass all those new values through the views
* Make num_of_components default to 3 everywhere and make it non-optional
Cumulative R^2/Q^2 values were added OPLSDA because they were provided, but aren't being rendered in the frontend.

Frontend changes:
* Give OPLSDA a num_of_compenents field to match PLSDA
* Render R^2 and Q^2 under the PC selector for both PLSDA and OPLSDA
* Rework the `watch` statement for number_of_components to properly cause the plot to re-render
* Default num_of_components to 3

### DIFF
--- a/web/src/components/vis/PlsdaPage/PlsdaPage.vue
+++ b/web/src/components/vis/PlsdaPage/PlsdaPage.vue
@@ -151,6 +151,7 @@ export default defineComponent({
               type="number"
               label="PC (X Axis)"
               min="1"
+              :max="controls.numComponents"
               outline="outline"
               :disabled="!controls.showScore && !controls.showLoadings"
             />
@@ -161,6 +162,7 @@ export default defineComponent({
               type="number"
               label="PC (Y Axis)"
               min="1"
+              :max="controls.numComponents"
               outline="outline"
               :disabled="!controls.showScore && !controls.showLoadings"
             />
--- a/web/src/components/vis/PlsdaPage/PlsdaPage.vue
+++ b/web/src/components/vis/PlsdaPage/PlsdaPage.vue
@@ -2,7 +2,7 @@
 import VisTileLarge from '@/components/vis/VisTileLarge.vue';
 import LayoutGrid from '@/components/LayoutGrid.vue';
 import {
-  computed, defineComponent, reactive, toRef, toRefs, watchEffect,
+  computed, ComputedRef, defineComponent, reactive, toRef, watch, watchEffect,
 } from '@vue/composition-api';
 import ScorePlot from './ScorePlot.vue';
 import LoadingsPlot from './LoadingsPlot.vue';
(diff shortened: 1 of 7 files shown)


web/src/components/vis/PlsdaPage/PlsdaPage.vue
```diff
@@ -53,20 +56,24 @@ export default defineComponent({
 
     watchEffect(() => {
       const pcX = Number.parseInt(controls.pcXval, 10);
-      if (!Number.isNaN(pcX)) {
+      if (!Number.isNaN(pcX) && pcX <= controls.numComponents) {
```
subdavis, line 59: Did we ever arrive at some universal rule about when to use `Number.isNaN` vs. `window.isNaN`?

You're using the result of parseInt here, so it's not a problem, but in general these methods produce different outputs.  `Number.isNaN('asdf') === false` for example.

I think typescript protects us from that sort of nonsense, so it doesn't matter here.

subdavis (changes requested):
Code looks good, one small issue.

The PC(x) and PC(y) selectors are able to go above num_components, at which point the graphs just don't update because there's no data above the index loaded from the server.  It defaults to 3, but above 3, the graphs just show pc 3.

The input boxes should probably be clamped to `(1, num_components)`

Screenshot from 2020-09-28 09-53-20

web/src/components/vis/PlsdaPage/PlsdaPage.vue
waxlamp: > You're using the result of parseInt here, so it's not a problem, but in general these methods produce different outputs. `Number.isNaN('asdf') === false` for example.

Oh goodness. TIL about this.

So, `window.isNaN(x)` seems roughly equivalent to `Number.isNaN(Number(x))`, is that right? That is, there are many values that convert to `NaN` when forced to a Number type, and `window.isNaN()` includes those semantics. If that's true, then I would say always use `Number.isNaN()`, and if you want the behavior of `window.isNaN()`, we should simply be explicit about it (by still using `Number.isNaN()` with an explicit conversion to `Number`).

But I literally just learned about this four minutes ago, so maybe there are corner cases or something.

dchiquit: `Number.isNaN()` looks better to me, so that's what I prefer. I too was not aware of the distinction, but that seems like a convincing argument to me.

subdavis: `Number.isNaN()` is the most correct, I agree.

dchiquit:
@subdavis I added a `max` to the text field, so you can no longer click the arrows to go above `numComponents`. The behavior where the plot does not update when an incorrect value is input is correct, like if a user types in '333' into the field accidentally.

subdavis (approved):
lgtm

jbeezley:
The R code looks good to me, though I think we should make an issue to allow json return types from opencpu as it is becoming almost comically inefficient without it.

dchiquit:
https://github.com/girder/viime/issues/675

PR #676, opened 2020-09-29
Add group selectors to OPLSDA

dchiquit:
Now arbitrary datasets can be used with OPLSDA, as long as they have two or more groups defined.

I found an example of doing the group filtering in R, but I don't know how to write/debug R so I implemented it in the python view instead.

### DIFF
--- a/viime/views.py
+++ b/viime/views.py
@@ -806,8 +806,10 @@ def get_oplsda(validated_table: ValidatedMetaboliteTable,
     group_column_name = groups.columns[0]
     # Initial page loads will not have groups selected yet, so default to the first two groups
     if group1 is None and group2 is None:
-        group1 = groups[group_column_name].unique()[0]
-        group2 = groups[group_column_name].unique()[1]
+        group_names = groups[group_column_name].unique()
+        group_names.sort()
+        group1 = group_names[0]
+        group2 = group_names[1]
 
     # Filter groups and measurements by the selected groups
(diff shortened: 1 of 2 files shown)


subdavis (changes requested):
Looks like the group menus initialize out of sync with oplsda groups.  

## Steps to reproduce

1. use anovaTest dataset from e2e
1. upload, go to oplsda, see below

Screenshot from 2020-09-30 17-03-15

It looks like both the first and second group have to be changed once for them to sync up, then they work.  If you change the second group first, they're still out of sync until you change the first one.

dchiquit:
@subdavis Fixed, the group name list was different on the server, so the first two groups (which it defaults to) were different.

subdavis (changes requested):
using the same dataset, when I first load the OPLSDA page, I still have this issue.  I pulled and rebuild the docker client image, then ran yarn serve to be sure.    Verified that your changes from 8ec3efaff2a3be8af69a6a7118d8cc7aa3c1bd6b are being built.

Screenshot from 2020-10-01 14-57-11

dchiquit:
I changed the behavior to just leave those fields blank when first loading the page.

subdavis (approved):
I don't think this behavior is ideal.  The drop-down selector does nothing when you change it (you have to change both), which will invariably be confusing to some users.  Maybe this can be addressed later.

PR #678, opened 2020-09-29
VIP Scores plot

dchiquit:
* Add `VipPlot.vue`
* Add VipPlot to OPLSDA and PLSDA pages
  * PLSDA page has a complete set of VIP scores for each "numComponent", so there is also a selector to choose which component to render
  * Both have a toggle to sort based on component name or on VIP score

### DIFF
--- a/web/src/components/vis/VipPlot.vue
+++ b/web/src/components/vis/VipPlot.vue
@@ -1,0 +1,215 @@
+<script lang="ts">
+import {
+  PropType, defineComponent, computed, ref, watchEffect, onMounted, Ref,
+} from '@vue/composition-api';
+import { axisBottom, axisLeft } from 'd3-axis';
+import { scaleBand, scaleLinear } from 'd3-scale';
+import { select, event } from 'd3-selection';
+import 'd3-transition';
+
+import useAxisPlot from './use/useAxisPlot';
+
+const margin = {
+  top: 20, right: 20, bottom: 100, left: 50,
+};
+const fadeInDuration = 500;
+const duration = 200;
+const maxLabelLength = 15;
+
+interface VipValue {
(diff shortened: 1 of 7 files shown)


web/src/components/vis/VipPlot.vue
```diff
@@ -0,0 +1,215 @@
...
+              tooltip.transition()
+                .duration(duration)
+                .style('opacity', 0.0);
+            }))
+          .transition()
+          .duration(fadeInDuration)
+          .attr('x', (d) => scaleX.value(d.col))
+          .attr('y', (d) => scaleY.value(d.vip))
+          .attr('width', scaleX.value.bandwidth())
+          .attr('height', (d) => dheight.value - scaleY.value(d.vip));
```
waxlamp, line 117: It looks like these same values are set on the enter selection, which makes it seem like these transitions aren't really needed. Do they trigger in some other situation?

waxlamp (approved):
Just had one question about transitions, but it shouldn't block merging of this PR.

web/src/components/vis/VipPlot.vue
dchiquit: This is what happened when I removed height:
image
You are correct that width is superfluous, as it doesn't change once the plot is loaded.

dchiquit: The transition happens when toggling the `Sort VIP plot values` toggle, so yes the transition is required. It's pretty satisfying to watch.

waxlamp: :+1:

subdavis (approved):
Tested, works great.

subdavis (approved):

PR #694, opened 2020-11-04
Add PLSDA to group predictions

dchiquit:
* Add a new endpoint, `/analyses/plsda_factors` which is used specifically for this visualization.
* Add a new analysis method `PLSDA` to the group predictions page.


subdavis (approved):
Tested, working.

PR #696, opened 2020-11-17
Use proper noun capitalization for "Viime"

waxlamp:
We should treat "Viime" as a name rather than an initialism. This PR changes the all-caps usage to proper noun capitalization in all but two instances--I've left it all-caps in the title property, and in the first heading, where the emphasis feels appropriate.


jeffbaumes (approved):
I'm good with this change.

PR #697, opened 2020-11-17
ROC Confidence Interval

dchiquit:


tests/test_analyses.py
```diff
@@ -44,11 +44,22 @@ def test_roc(client, test_dataset):
...
     assert resp.status_code == 200
-    keys = {'sensitivities', 'thresholds', 'specificities', 'auc'}
+    keys = {
+        'sensitivities',
+        'thresholds',
+        'specificities',
+        'auc',
+        'lower_bound',
+        'upper_bound',
+        'y'
```
waxlamp, line 54: `y` is much less descriptive than the other values. Is there a better name we could use?

waxlamp (approved):

tests/test_analyses.py
dchiquit: It's provided from the R library, so not really. It's also not used on the frontend, so I don't think it's very significant. It can be removed from the python code that calls the R analysis if you think it's misleading.

PR #698, opened 2020-11-18
Add OPLSDA to the group prediction analysis list

dchiquit:
Add the OPLSDA analysis option to the existing Group Prediction analyses.
* The Metabolite Source is disabled, as OPLSDA only returns one component (PC0).
* While OPLSDA is selected, changing the groups will also rerun the analysis (by calling the method `getFactors()`). OPLSDA is unique in that it requires two groups as arguments, so I just recycled the group selectors that are already being used to perform the ROC analysis.

The python code is basically copied from PLSDA, but since there is only ever 1 factor returned, the loop over every factor is not necessary.

### DIFF
--- a/viime/views.py
+++ b/viime/views.py
@@ -1033,11 +1033,48 @@ def get_plsda_factors(validated_table: ValidatedMetaboliteTable,
         sorted_scores = sorted(filtered_scores, key=lambda s: s[1])
         metabolites += [column_names[j] for j, score in sorted_scores]
         factors += [i] * len(sorted_scores)
+    return jsonify({
+        'factor': factors,
+        'metabolites': metabolites,
+        'max_vip': 3,  # TODO: max_vip
+    })
+
+
+@csv_bp.route('/csv/<uuid:csv_id>/analyses/oplsda_factors', methods=['GET'])
+@use_kwargs({
+    'num_of_components': fields.Integer(missing=3),
+    'threshold': fields.Float(missing=0.4),
+    'group1': fields.String(required=True),
+    'group2': fields.String(required=True),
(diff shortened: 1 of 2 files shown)


subdavis:
I can't figure out how to use this.  I've clicked all the combinations of options I can think of.

Screenshot from 2020-11-20 14-07-40

I get a 400 for everything from `roc` endpoint,

Screenshot from 2020-11-20 14-10-29

and the new `oplsda_factors` endpoint says:

```
'y' factor length must be equal to the number of rows of 'x'
```

Am I doing something wrong?

dchiquit:
@subdavis I added a new view to calculate the OPLSDA factors, but the Netlify preview is aimed at the production API backend, so that new view is not available. In your network tab you should see 404s whenever the page tries to load `oplsda_factors`. I think the only way to see this in action is to pull down the branch and run it locally.

subdavis:
I have been running it locally.  rebuilt both backend and opencpu.

dchiquit:
I tested it with `Breastcancer.csv` and it worked, but `Cancer_chemo_muscle_NMR.csv` didn't work for me either. I will look closer.

dchiquit:
@subdavis fixed

subdavis (approved):
Tested working now.

FYI, this bug is also present when the new OPLSDA option is selected: This also affects this PR: https://github.com/girder/viime/issues/699

When you change the group, for example.

1 request is fired when you change group, and when the results of `oplsda_factors` returns, a second request is sent.  These will race to determine the end state, though most likely the second (correct) result will win.

Screenshot from 2020-11-23 11-19-33

PR #705, opened 2021-01-27
Fix SVG downloads not triggering

### LINKED ISSUE #703: Downloading PNG not working for BoxPlots and Group Prediction (AUC)
Hello,

the PNG image download for Box Plots (single or multiple?) does not work.
Same is also true for Group Prediction (AUC) downloads.

Jobs: 
https://viime.org/#/app/pretreatment/3bd5b9cd-fc6a-46ca-9798-62b08c5dda02/analyze/boxplot
https://viime.org/#/app/pretreatment/3bd5b9cd-fc6a-46ca-9798-62b08c5dda02/analyze/roc

Thanks,
Biswa

box_plots
roc_auc

dchiquit:
`el` is a template ref that allows the `VisTileLarge` component's TypeScript to interact directly with the mounted DOM element. Both the `ref="el"` on the template and the `const el = ref(...);` were set up correctly, but the ref `el` was not returned from the `setup` function, so Vue did not know to autowire that ref when the template was mounted. The download was searching for an `<svg>` element inside the placeholder value of the ref (an empty div), and when it could not find an SVG, it simply did nothing.

The solution is return `el` from the setup function.

Fixes #703

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ setup(
         'sqlalchemy-utils',
         'webargs >=5.5.3, <6',
         'Werkzeug>=0.15',
-        'xlrd>= 1.0.0'
+        'openpyxl'
     ],
     extras_require={
         'memcached': ['pylibmc'],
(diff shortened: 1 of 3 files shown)


jeffbaumes (approved):
Looks good, thanks!

